PR #36, opened 2021-01-05
[MYEN-511] add handling rejection message

dwojno:


manihagh (approved):

src/dgraph/dgraph.service.ts
```diff
@@ -37,6 +37,7 @@ export class DgraphService {
         parentNamespace
         isAccepted
         createdAt
+        isRejected
```
jrhender, line 40: Perhaps not worth changing at this point but having both the `isAccepted` and `isRejected` fields means that we could end up with claims that are both accepted and rejected and so perhaps it would be better to have a `approvalStatus` field instead or something similar. A bit of a pain to implement now as we would have to migrate existing data. Maybe there's another way to constrain these fields so that they are mutually exclusive?

jrhender (commented):
Doesn't the `ClaimIssue` class need to be modified as well?

jrhender (approved):
Nice work!

PR #38, opened 2021-01-07
[MYEN-521] fix namespace search implementation

dwojno:


manihagh (approved):

src/namespace/namespace.service.ts
```diff
@@ -53,15 +54,68 @@ export class NamespaceService {
...
     text: string,
-  ): Promise<(Application | Organization)[]> {
-    const res = await this.dgraph.query(`{
-       data(func: match(namespace, "${text}", 32)) @filter(eq(dgraph.type, ["App", "Org"])) {
+    type?: NamespaceEntities,
+  ): Promise<(Application | Organization | Role)[]> {
+    const directNamespacesPromise = this.dgraph.query(
+      `{
+        data(
+          func: ${type ? `type(${type})` : 'has(namespace)'})
```
jrhender, line 62: Is this `'has(namespace)'` meant to capture the `Role`, `App`, `Org` DGraph types? If so, is it possible to more explicitly list the types in this function?
E.g.  ```func: ${type ? `type(${type})` : 'type(App) OR type(Role) OR type(Org)```
Doesn't seem to possible from the Dgraph docs, but maybe you know a way

jrhender (approved):
Looks good!

src/namespace/namespace.service.ts
dwojno: Nope this is not possible, we can do this in @filter but not in the func, but still func needs to be there :)

PR #39, opened 2021-01-10
Optionally disabling did and ens sync using config

jrhender:
This is a small quality of life improvement as I found myself occasionally disabling sync while developing.
Enabled by default so as to not affect prod.


manihagh (approved):

dwojno (commented):
Please add both variables to .env files with default value 😄

jrhender:
> @jrhender Please add both variables to .env files with default value 😄 please remember that we now have CI/CD so every change in .env files needs to reflected in travis settings :)

@dwojno I don't think it's necessary to add to Travis settings if there is a value in the .env file, right?

dwojno (approved):

PR #40, opened 2021-01-11
Updating restart policy to "unless-stopped" in prod

jrhender:
Removed limit on number of restarts as not allowed by unless-stopped policy


manihagh (approved):

dwojno:
@jrhender please add how many attempts it may restart the container, to prevent endless loop of restarting and killing the vm :)

PR #42, opened 2021-01-11
[MYEN-523] fix creating not connected definitions

dwojno:


manihagh (approved):
LGTM, thanks

manihagh (approved):

src/application/ApplicationTypes.ts
```diff
@@ -18,6 +18,7 @@ export interface Application extends DGraphObject {
  * Interface describing generic Application's Definition model
  */
 export interface AppDefinition extends Definition {
+  uid?: string;
```
jrhender, line 21: @dwojno Why not extend `DGraphObject` interface as well here? Is it because you don't want to include the `dgraph.type` property?

jrhender: It looks like `readonly 'dgraph.type' = 'AppDefinition';` is already in `ApplicationDefinitionDTO` class, so should be okay to have `AppDefinition` extend `DGraphObject`

src/dgraph/dgraph.service.ts
```diff
@@ -178,7 +180,7 @@ export class DgraphService {
    */
   public async mutate(data: unknown) {
     const instance = await this.getInstance();
-    const txn = await instance.newTxn();
+    const txn = instance.newTxn();
```
jrhender, line 183: 👍

src/application/ApplicationTypes.ts
dwojno: Yeah great idea, I'll extend all from `DGraphObject' 👍

dwojno: @jrhender I've updated the PR :)

jrhender (approved):
Looks good

docker/prod/Dockerfile
```diff
@@ -27,4 +27,3 @@ RUN chmod +x /app/wait-for-db.sh
 
 ENV NODE_ENV=production
 
-USER node
```
jrhender, line 30: @dwojno  Is this a temporary removal for the fix or is this permanent?

dwojno: This only temporary, cause there were an issue to write to file while being an user that is not a root ;)

PR #44, opened 2021-01-12
[MYEN-524] remove posibility to delete anything through CI/CD

dwojno:
As @jrhender pointed out we have an `access denied` error while removing old db dumps running in CI/CD.
So I was thinking that this is not a bad idea ;) we shouldn't be able to remove that important data while using CI/CD. We still have a cron job that is doing that every 12 hours, so I've added a script to backup db without removing it from the server in CI/CD env


manihagh (approved):

jrhender (approved):
Looks good 👍

PR #45, opened 2021-01-12
[MYEN-525] update iam cache server search implementation

dwojno:


src/namespace/namespace.service.ts
```diff
@@ -54,15 +54,31 @@ export class NamespaceService {
...
+    const { typeDefinitionFilters, typeFilters } = (types || []).reduce(
+      (filters, type, index) => {
+        if (index > 0) {
+          filters.typeFilters += ` OR type(${type})`;
+          filters.typeDefinitionFilters += ` OR type(${type}Definition)`;
+          return filters;
+        }
+        filters.typeFilters += `type(${type})`;
+        filters.typeDefinitionFilters += `type(${type}Definition)`;
+        return filters;
```
jrhender, line 68: I think could change to something like this to avoid duplication of `type(${type})` and `type(${type}Definition)`. Kind of minor but avoids potential future bug.
```
if (index > 0) {
  filters.typeFilters += ` OR `;
  filters.typeDefinitionFilters += ` OR `;
}
filters.typeFilters += `type(${type})`;
filters.typeDefinitionFilters += `type(${type}Definition)`;
return filters;
```

Or maybe
```
const typeFilter = `type(${type})`
const typeDefFilter = `type(${type}Definition)`
if (index > 0) {
  filters.typeFilters += ` OR ${typeFilter}`;
  filters.typeDefinitionFilters += ` OR ${typeDefFilter}`;
  return filters;
}
filters.typeFilters += typeFilter;
filters.typeDefinitionFilters += typeDefFilter;
return filters;
```

jrhender (approved):
Looks good

src/namespace/namespace.service.ts
dwojno: 👍

PR #50, opened 2021-01-14
[MYEN-535] optimize ENS sync process

dwojno:


src/ENS/ens.service.ts
```diff
@@ -128,21 +123,8 @@ export class EnsService {
         return this.publicResolver.name(node);
       }),
     );
-    const uniqDomains: Record<string, unknown> = {};
-    for (const item of domains) {
-      if (item && item.endsWith(domain) && !uniqDomains[item]) {
```
jrhender, line 133: @dwojno Because we're not filtering for subdomains anymore, will this change cache all of ENS?

dwojno: @jrhender nope, we're only getting domains that triggered event TextChanged and the text fields is 'metadata', so as far as I know, only iam-client-lib is adding this kind of text property to the ENS domain

jrhender (approved):
Nice, I tested it out, it is wayyyyyy faster now!

PR #54, opened 2021-01-19
Changing CommonJS imports to ES Module syntax

jrhender:
The addition of  `"esModuleInterop": true` means that we should import CommonJS modules in compliance with es6 modules spec. This is currently causing bug in where full claims are not returned when retrieving DID Document.


jrhender:
@manihagh @dwojno Feel free to merge after review as would be good to get this deployed as this is a currently causing a bug where IPFS claim data isn't included in returned DID Document.

dwojno (approved):

PR #55, opened 2021-01-20
[MYEN-567] update org endpoints to return suborgs

dwojno:


src/organization/OrganizationDTO.ts
```diff
@@ -140,5 +140,7 @@ export class OrganizationDTO implements Organization {
   @IsOptional()
   parentOrg?: Organization;
 
+  subOrgs?: Organization[];
```
jrhender, line 143: Should this have the `@IsOptional()` tag as well?

dwojno: 👍

jrhender (approved):
LGTM

PR #56, opened 2021-01-22
[MYEN-371] add authentication for iam cache server

dwojno:
- [x] add login controller and use login strategy from passport-did-auth
- [x] update docker config to generate asymmetric RSA key pair
- [x] add guards for all routes using jwt strategy to only pass user having valid jwt token
- [x] use asymmetric token validation

UPDATE:

- [x] add option to disable auth


src/main.ts
```diff
@@ -11,7 +11,12 @@ async function bootstrap() {
     .setTitle('API')
     .setDescription('Cache Server API documentation')
     .setVersion('1.0')
+    .addBearerAuth(
+      { type: 'http', scheme: 'bearer', bearerFormat: 'JWT' },
+      'access-token',
+    )
     .build();
+
   const document = SwaggerModule.createDocument(app, options);
```
jrhender, line 20: Should we be disabling the "try it out" functionality in Swagger as it won't work now that we have auth 🤔 ?

jrhender (commented):
Looks really good, left a few comments though. A bunch are optional rename suggestions 😄 

Another thing is whether or not the cache-server's documentation needs to be updated? Maybe something in the README and index.md like "before calling endpoints, add `/login` to get token and then set in `Auth` header"

src/main.ts
dwojno: On the upper right corner you have an option to auth using jwt token :)

jrhender: Oh, cool, I didn't know Swagger had that. So the steps to use Swagger would be:
1. Generate "IdentityProof" claim
2. Call login CacheServer endpoint (maybe with Swagger) and get back jwt
3. Use jwt in subsequent Swagger calls

How would someone wanting to use Swagger accomplish step 1? Should we give them a script that can use? Or maybe we could provide some endpoint/app/dapp that generates "IdentityProof" claim based on a private key or some connected wallet 🤔 ?

dwojno: Yeah I agree we need to make the step one easier, my first idea was to find a way to run some script on swagger ui load, to log in using metamask and generate token, but I can't find a way to do this using nest swagger module

PR #62, opened 2021-02-04
feat(auth): protect deleting endpoint

dwojno:
Only logged user which is a requester for certain not accepted and not rejected claim can cancel it.


jrhender (approved):
Looks good to me 👍

PR #65, opened 2021-02-09
feat(cookies): add option to store token in cookies

dwojno:


src/auth/jwt.strategy.ts
```diff
@@ -3,12 +3,21 @@ import { PassportStrategy } from '@nestjs/passport';
...
+import { Request } from 'express';
 
 @Injectable()
 export class JwtStrategy extends PassportStrategy(Strategy) {
   constructor(configService: ConfigService) {
     super({
-      jwtFromRequest: ExtractJwt.fromAuthHeaderAsBearerToken(),
+      jwtFromRequest: ExtractJwt.fromExtractors([
+        ExtractJwt.fromAuthHeaderAsBearerToken(),
+        (req: Request) => {
```
jrhender, line 14: Nice, I guess the multipleExtractors means that this is backwards compatible with `iam-client-lib` versions that were sending the jwt as a bearerToken?

jrhender (approved):

src/auth/jwt.strategy.ts
dwojno: Yes it is, that was the main goal :)

PR #66, opened 2021-02-11
feat(subOrgs): get org subOrgs hierarchy

dwojno:


kim-energyweb (approved):
LGTM

src/organization/organization.service.ts
```diff
@@ -299,4 +300,58 @@ export class OrganizationService {
...
+          subOrgs: await this.getDefinitionForOrgs(org.subOrgs),
+        })),
+      )
+    );
+  }
+
+  public async getOrganizationNestedSubOrgs(namespace: string) {
+    const res = await this.dgraph.query(`
+    {
+      orgs(func: eq(namespace, "${namespace}")) @filter(type(Org)) @recurse(depth: 20, loop: true) {
```
jrhender, line 340: Cool, I didn't know about this `@recurse` functionality. Why the need to loop? I would think there wouldn't be loops in the organization hierarchies

src/organization/organization.service.ts
```diff
@@ -299,4 +300,58 @@ export class OrganizationService {
...
+    return definition;
+  }
+
+  private getDefinitionForOrgs(orgs?: Organization[]) {
+    return (
+      orgs &&
+      Promise.all(
+        orgs.map(async org => ({
+          ...org,
+          definition: await this.getOrgDefinitionByUid(org.definition.uid),
```
jrhender, line 330: Why do you need to get the OrgDefinition again? It seems like a bunch of the data (ie. uid, name, owner, namespace, definition) is being retrieved already in `getOrganizationNestedSubOrgs`?

src/organization/organization.service.ts
```diff
@@ -299,4 +300,58 @@ export class OrganizationService {
     };
     return json?.Data?.[0]?.['subOrgs'] || [];
   }
+
+  private async getOrgDefinitionByUid(uid: string) {
```
jrhender, line 304: Minor thing, but would be nice to have private methods after public in the file

jrhender (approved):
Left a few comments/questions, but looks good. The dgraph syntax is tricky! 😆

src/organization/organization.service.ts
dwojno: In the @recurse function we can get only top level props, so the definition looks like this { uuid: string } and we need to get the rest manually :)

src/organization/organization.service.ts
dwojno: 👍

src/organization/organization.service.ts
dwojno: If the loop would be `false` the dgraph will stop looking when found the same node as earlier :) so this is just to be sure to find all nested suborgs in whatever configuration :)

PR #67, opened 2021-02-12
fix(cookies): samesite none cookie issue

dwojno:
https://medium.com/@arcagarwal/same-site-changes-in-chrome-1c86973454f9

### DIFF
--- a/src/auth/login.controller.ts
+++ b/src/auth/login.controller.ts
@@ -1,7 +1,8 @@
 import { Controller, Post, Req, Res, UseGuards } from '@nestjs/common';
 import { LoginGuard } from './login.guard';
-import { Request, Response } from 'express';
+import { CookieOptions, Request, Response } from 'express';
 import { ApiBody } from '@nestjs/swagger';
+import useragent from 'useragent';
 
 @Controller()
 export class LoginController {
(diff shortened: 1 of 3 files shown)


src/auth/login.controller.ts
```diff
@@ -19,9 +20,15 @@ export class LoginController {
   })
   @Post('login')
   async login(@Req() req: Request, @Res() res: Response) {
-    res.cookie('token', req.user, {
+    const cookiesOptions: CookieOptions = {
       httpOnly: true,
-    });
+    };
+    const { family, major } = useragent.parse(req.headers['user-agent']) || {};
+    if (family === 'Chrome' && +major > 61) {
```
jrhender, line 27: The article mentions that this should be for chrome versions greater than 66, so why the 61? https://www.chromium.org/updates/same-site/incompatible-clients

Also, the root problem here is that the default `sameSite` policy is "lax", right? I think it is becoming a browser standard now to have the default policy be "lax" (e.g. https://hacks.mozilla.org/2020/08/changes-to-samesite-cookie-behavior/) so maybe we should just be setting it to "none" for all browsers?

dwojno: Sorry for that my mistake, yeah you're totally right it's 66 👍 
So maybe let's add by default sameSite=none, and remove this property on chrome from 51 till 66?
@jrhender what do you think?

dwojno: @jrhender I've updated the code

manihagh (approved):

PR #70, opened 2021-02-16
feat(auth): add refresh token implementation

dwojno:


manihagh (approved):

src/auth/login.controller.ts
```diff
@@ -20,17 +36,81 @@ export class LoginController {
...
+  @Get('refresh_token')
+  async refreshToken(@Req() req: Request, @Res() res: Response) {
+    const refreshTokenString =
+      req.cookies[this.configService.get<string>('JWT_REFRESH_TOKEN_NAME')];
+
+    if (!refreshTokenString) {
+      throw new UnauthorizedException();
+    }
+
+    const { userDid, id, verifiedRoles } =
```
jrhender, line 82: This is the token id right? If so, maybe nice to have it renamed to `tokenId`

src/auth/login.controller.ts
```diff
@@ -20,17 +36,81 @@ export class LoginController {
...
+
+    const cookiesOptions = this.cookiesServices.getCookiesOptionBasedOnUserAgent(
+      req.headers['user-agent'],
+    );
+
+    const [token, refreshToken] = await Promise.all([
+      this.tokenService.generateAccessToken({ did: userDid, verifiedRoles }),
+      this.tokenService.generateRefreshToken({
+        userDid,
+        verifiedRoles,
```
jrhender, line 97: Should we be reusing the same `verifiedRoles` that we received from the old refresh token? I wonder if it might make more sense to refresh the roles so that the accessToken is up to date? If so, maybe we don't need to store the roles in the refreshToken

src/auth/login.controller.ts
dwojno: 👍

src/auth/login.controller.ts
dwojno: 👍

jrhender (approved):
Nice work 👍

PR #73, opened 2021-02-23
feat(didSync): add concurrency limit to refresh processing

jrhender:
I don't have any great evidence for it but I think it is possible that DGraph locking up could be the cause of the iam-cache-server going down. To reduce what might be a cause of strain on DGraph, I am thinking that it might be a good idea to limit the number of DidDoc refreshes that occur concurrently so that DGraph (and the blockchain RPC) it only given a new request once it is done processing a previous one. Do you think this is a good idea?

### DIFF
--- a/src/did/did.processor.ts
+++ b/src/did/did.processor.ts
@@ -8,21 +8,27 @@ import { DID } from './DidTypes';
 export class DIDProcessor {
   private readonly logger: Logger;
 
-  constructor(private readonly didService: DIDService) {
+  constructor(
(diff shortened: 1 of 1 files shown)


dwojno (approved):

jrhender:
Will finalize and merge this once the auth changes have been merged.

Harasz:
@jrhender What is the status of this PR?

Harasz:
I am closing the thread due to inactivity.

PR #75, opened 2021-02-25
feat(logging): add loggin to files and sentry integration

dwojno:


src/interceptors/not-found.interceptor.ts
```diff
@@ -13,12 +14,11 @@ export class NotFoundInterceptor implements NestInterceptor {
     {
       return next.handle().pipe(
         tap(data => {
-          const notFoundHTTPcode = 404;
           if (data === null)
             context
               .switchToHttp()
               .getResponse()
-              .status(notFoundHTTPcode);
+              .status(HttpStatus.NOT_FOUND);
```
jrhender, line 21: 👍

package.json
```diff
@@ -68,12 +70,15 @@
     "swagger-ui-express": "^4.1.4",
     "typechain": "^2.0.0",
     "useragent": "^2.3.0",
-    "uuid": "^8.3.2"
+    "uuid": "^8.3.2",
+    "winston": "^3.3.3",
```
jrhender, line 74: I haven't seen/used winston before but it looks good 👍

src/logger/logger.service.ts
```diff
@@ -0,0 +1,95 @@
...
+
+@Injectable({ scope: Scope.TRANSIENT })
+export class Logger extends NestLogger implements LoggerService {
+  public readonly logger: WinstonLogger;
+  constructor(
+    private readonly configService: ConfigService,
+    private readonly sentryService: SentryService,
+  ) {
+    super();
+    const isProduction = configService.get<string>('NODE_ENV') === 'production';
```
jrhender, line 25: This is to differentiate "dev" from deployed environments, right? So both "dev" and "volta" cache-servers would have "production" logging? Am I understanding correctly?

jrhender (commented):
Looks good. I think it would be good to add to README that this app is using Sentry and a note on how to configure/disable for dev environment. What do you think?

src/logger/logger.service.ts
dwojno: Yup both of them will use production env, "notProduction" env is just for our local development environment :)

jrhender (approved):

PR #89, opened 2021-03-17
feat(graphQL): add graphql api

dwojno:


src/main.ts
```diff
@@ -25,7 +25,12 @@ async function bootstrap() {
   app.useGlobalPipes(new ValidationPipe());
 
   // add security middleware
-  app.use(helmet());
+  app.use(
+    helmet({
+      contentSecurityPolicy:
+        process.env.NODE_ENV === 'production' ? undefined : false,
```
jrhender, line 31: Why the need for a difference between dev and production here? 
And why is the contentSecurityPolicy `undefined` for production?

src/modules/claim/claim.service.ts
```diff
@@ -147,33 +160,82 @@ export class ClaimService {
...
+      });
+    }
+
+    if (user) {
+      qb.andWhere(new Brackets(query => {
+        query.where(':user = ANY ("claimIssuer")', {
+          user,
+        }).orWhere(':user = requester', { user })
+      }))
+    }
```
jrhender, line 197: What is the difference between `user` and `did`? The logic seems to be duplicated

src/main.ts
dwojno: Yeah, I'm going to change this for false, for all of env's, cause this is needed for graphql playground to be able to run.

src/modules/claim/claim.controller.ts
```diff
@@ -41,14 +40,9 @@ export class ClaimController {
...
     private readonly claimService: ClaimService,
     private readonly nats: NatsService,
-    @InjectQueue('claims') private claimQueue: Queue<string>,
     private readonly logger: Logger,
   ) {
     this.logger.setContext(ClaimController.name);
-    this.nats.connection.subscribe(`*.${NATS_EXCHANGE_TOPIC}`, async data => {
-      this.logger.debug(`Got message ${data}`);
-      await this.claimQueue.add('save', data);
-    });
```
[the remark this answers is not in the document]
dwojno: For me controller needs to have as little as possible responsibilities (validation and passing data to controlled services), all business logic should happen is services :)

src/modules/claim/claim.service.ts
dwojno: I can change the user to `currentUser`, to make it more clear what is this `user`

jrhender: @dwojno Ok, I think this rename would be a good idea

jrhender (approved):
Looks good

PR #92, opened 2021-03-23
feat(assets): add assets caching

dwojno:


src/modules/did/did.service.ts
```diff
@@ -99,49 +109,124 @@ export class DIDService {
...
+   * Gets all logs for a given DID
+   * @param documentEntity
+   */
+  private async getAllLogs(id: string): Promise<IDIDLogData> {
+    const genesisBlockNumber = 0;
+    const readFromBlock = utils.bigNumberify(genesisBlockNumber);
+    return this.resolver.readFromBlock(id, readFromBlock);
+  }
+
+  private updateServices(
```
jrhender, line 162: Maybe a name like `getMissingClaims` would be more accurate as an `IClaim[]` is being returned and there is no update that is taking place (i.e. no data is actually mutated)

src/modules/did/did.service.ts
```diff
@@ -99,49 +109,124 @@ export class DIDService {
...
-      await this.refreshCachedDocument(did);
-      cachedDIDDocument = await this.didRepository.findOne(did.id);
-    }
-    const didEntity = new DIDDocumentEntity(did, cachedDIDDocument);
-    const resolvedDocument = await didEntity.getResolvedDIDDocument();
-    if (enhanceWithClaims) {
-      this.enhanceWithClaims(didEntity, resolvedDocument);
-    }
-    return resolvedDocument;
+  public async getById(did: string): Promise<IDIDDocument> {
```
jrhender, line 194: A bit of a personal preference, but I would put public methods before private in the file

src/modules/did/did.processor.ts
```diff
@@ -17,8 +16,7 @@ export class DIDProcessor {
   public async processDIDDocumentUpsert(job: Job<string>) {
```
jrhender, line 16: I think ideally the naming would be consistent, so if the didService method being called is `addCacheDocument`, then maybe this method should be called `processDIDDocumentAddition`

src/modules/did/did.entity.ts
```diff
@@ -1,19 +1,177 @@
...
+  serviceEndpoint: string;
+
+  @Field()
+  type: string;
+
+  @Field(() => Int)
+  validity: any;
+
+  [key: string]: string | number;
+}
```
jrhender, line 129: Interesting that all of the fields of the DIDDocument have been broken out. I'm on the fence as to whether or not this is a good idea. On one hand, it would be possible performing richer queries on the data. On the other hand, it adds to the complexity of the caching. I think if you've already implemented it we can use it but I'm wondering what are your thoughts are on why you decided to implement all of the DIDDocument properties?

src/modules/did/did.entity.ts
```diff
@@ -1,19 +1,177 @@
...
 
 @Entity()
-export class DIDEntity {
+@ObjectType()
+export class DIDEntity implements IDIDDocument {
```
jrhender, line 133: Could we rename this to `DIDDocumentEntity` plz? I think it's what I should have named it initially 😅

src/modules/did/did.service.ts
```diff
@@ -11,47 +11,57 @@ import { ConfigService } from '@nestjs/config';
...
   constructor(
     private readonly config: ConfigService,
     private readonly schedulerRegistry: SchedulerRegistry,
-    private readonly resolverFactory: ResolverFactory,
     private readonly httpService: HttpService,
     @InjectQueue('dids') private readonly didQueue: Queue<string>,
     private readonly logger: Logger,
     @InjectRepository(DIDEntity)
     private readonly didRepository: Repository<DIDEntity>,
+    private readonly provider: Provider,
```
jrhender, line 48: Nice, I like this refactor into the common provider 👍

src/modules/assets/assets.entity.ts
```diff
@@ -0,0 +1,37 @@
...
+  offeredTo?: string;
+
+  @Column()
+  @Field()
+  timestamp: string;
+
+  @OneToOne(
+    () => DIDEntity,
+    did => did.id,
+    { eager: true },
```
jrhender, line 32: Why this `eager: true` here?

src/modules/assets/assets.entity.ts
```diff
@@ -0,0 +1,37 @@
...
+  )
+  @JoinColumn()
+  @Field(() => DIDEntity)
+  document: DIDEntity;
```
jrhender, line 36: I think this approach of having the Asset being composed fo the DIDEntity 👍 
So in postgres, Assets and DID(Documents) will each be a separate table?

jrhender (commented):
Looks good so far. A few minor comments.

src/modules/assets/assets.service.ts
```diff
@@ -0,0 +1,129 @@
...
+import { Provider } from '../../common/provider';
+import { IdentityManager } from '../../ethers/IdentityManager';
+import { IdentityManagerFactory } from '../../ethers/IdentityManagerFactory';
+import { OfferableIdentityFactory } from '../../ethers/OfferableIdentityFactory';
+import { DIDService } from '../did/did.service';
+import { AssetDto } from './assets.dto';
+import { Asset } from './assets.entity';
+
+@Injectable()
+export class AssetsService {
```
jrhender, line 14: @dwojno Could have some sync of assets? Is that possible to implement? I guess we need to go through the logs of the `IdentityManager`?

src/modules/did/did.service.ts
```diff
@@ -99,49 +109,124 @@ export class DIDService {
...
-    return resolvedDocument;
+  public async getById(did: string): Promise<IDIDDocument> {
+    const cachedDIDDocument = await this.didRepository.findOne(did);
+    if (cachedDIDDocument) return cachedDIDDocument;
+
+    this.logger.info(
+      `Requested document for did: ${did} not cached. Queuing cache request.`,
+    );
+
+    // awaiting refresh so that subsequent calls don't duplicate cached DID Document
```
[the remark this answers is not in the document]
dwojno: 👍

src/modules/ens/ens.service.ts
```diff
@@ -65,7 +64,7 @@ export class EnsService {
       );
       this.schedulerRegistry.addInterval('ENS Sync', interval);
     }
-
+    // this.syncENS();
```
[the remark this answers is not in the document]
dwojno: Yeah, sorry this should be there ;)

src/modules/did/did.service.ts
```diff
@@ -99,49 +109,124 @@ export class DIDService {
...
   }
 
   /**
-   * Insert or update the DID Document cache for a given DID.
+   * Adds the DID Document cache for a given DID.
    * Also retrieves all claims from IPFS for the document.
    * @param {DID} did
    */
-  public async upsertCachedDocument(did: DID): Promise<void> {
+  public async addCachedDocument(did: string) {
```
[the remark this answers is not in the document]
dwojno: Nope :) the primary key of each DIDEntity is the did, so it will just basically update the entity :)

src/modules/did/did.service.ts
dwojno: 👍

src/modules/did/did.service.ts
dwojno: 👍

src/modules/did/did.processor.ts
dwojno: 👍

src/modules/did/did.entity.ts
dwojno: For purpose of the caching is not adding any complexity, cause we're storing almost all of this data as `jsonb` column and it will accept any kind of data, this is made just for graphql queries, so we'll be able to access deeply nested objects.

src/modules/did/did.entity.ts
dwojno: :) no problem :)

src/modules/assets/assets.entity.ts
dwojno: I thought it would be better to automatically resolve this relation, cause we always want to see the didDoc of the asset.

src/modules/assets/assets.entity.ts
dwojno: Yes, Asset and DIDDoc are seperate tables in postgres and asset has onetoone relation with didDoc :)

src/modules/assets/assets.service.ts
dwojno: Yes of course, currently I'm working on this :)

src/modules/assets/assets.entity.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  @Column()
+  @Field()
+  timestamp: string;
+
+  @ManyToOne(
+    () => Asset,
+    asset => asset.history,
+  )
+  asset?: Asset;
```
jrhender, line 96: In what cases would `AssetHistory` not have an associated asset?

jrhender (approved):
Looks good. A few questions for you if you have time.
Also, could you add a comment or point out the location of the field that can only be determined from direct listening of events but not from sync from chain?

src/graphql/config.ts
```diff
@@ -7,7 +7,7 @@ export const getGraphQlConfig = (
   const isProduction = configService.get<string>('NODE_ENV') === 'production';
   return {
     playground: true,
-    // Generate graphql schema for dev to store it repo
+    // Generate graphql schema for dev to store it in repo
     autoSchemaFile: isProduction ? true : 'src/graphql/schema.gql',
```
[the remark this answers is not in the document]
dwojno: This is nothing to do with performance :) Basically i though that it would be great to save the schema in the repo, so we or other members of the team would easily check it out :) and of course we'll be able to easier track the changes

src/graphql/schema.gql
```diff
@@ -92,7 +92,7 @@ type AuthenticationSchema {
   block: Int
   publicKey: String!
   type: String!
-  validity: Int!
+  validity: JSONObject!
```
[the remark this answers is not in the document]
dwojno: Validity is a BigNumber or some other 'special' type base on object so to not make schema for every field I've added this general JSON schema :)

src/migrations/1617022027299-AddAssets.ts
```diff
@@ -0,0 +1,44 @@
+import {MigrationInterface, QueryRunner} from "typeorm";
+
+export class AddAssets1617022027299 implements MigrationInterface {
```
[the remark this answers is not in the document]
dwojno: Yes :) we've got a npm script to do that :) `npm run migration:generate $MIGRATION_NAME`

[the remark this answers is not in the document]
dwojno: Migration are runned every time app starts :) but still we need to generate them after any change made to entities :)

src/modules/assets/assets.dto.ts
```diff
@@ -0,0 +1,28 @@
...
+  }
+
+  @IsString()
+  id: string;
+
+  @IsString()
+  owner: string;
+
+  @IsNumber()
+  at: number;
```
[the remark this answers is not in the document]
dwojno: `at` is the time from chain when the event happened :) it's the number of seconds since `January 01, 1970, 00:00:00` :) so to get JS time we need to multiply this value by `1000` to get milliseconds since `since January 01, 1970, 00:00:00` :)

src/modules/assets/assets.entity.ts
dwojno: Normally we won't have the `assetId` field in the entity, cause it is created by the typeorm to keep the foreign key. But this has a couple of downsides, to create a relation you need to pass the whole asset entity, even if you already have the assetId and you don't want to perform a request to db to get it. Second downside is that you're not able to filter the entity by the assetId field, cause TypeOrm doesnt not this field. So to overcome this I've added a special field named `assetId` and now we're able to create relationship with asset just be passing the `assetId` and we're able to filter by this value :)

dwojno: I've also added a typechecker to check if one of this fields passed to entity creator:
```ts
Omit<AssetsHistory, 'id'> & ({ asset: Asset } | { assetId: string })
```

src/modules/assets/assets.dto.ts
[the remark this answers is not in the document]
jrhender: @dwojno Thanks, but why is the property on the `AssetDto`? Is `AssetDto` really an `AssetEventDto`?

src/modules/assets/assets.entity.ts
jrhender: Ok, thanks

PR #100, opened 2021-04-01
fix(role.service): check for empty conditions array

jrhender:
Same fix that was made here: https://github.com/energywebfoundation/iam-client-lib/commit/560f85a801a889469f430b34e91576cff2735202

### DIFF
--- a/src/modules/role/role.service.ts
+++ b/src/modules/role/role.service.ts
@@ -160,7 +160,7 @@ export class RoleService {
 
     if (!enrolmentPreconditions || enrolmentPreconditions.length < 1) return;
     for (const { type, conditions } of enrolmentPreconditions) {
-      if (type === 'role' && conditions) {
+      if (type === 'role' && conditions && conditions?.length > 0) {
         const conditionMet = didDocument.service.some(
           ({ claimType }) =>
             claimType && conditions.includes(claimType as string),


src/modules/role/role.service.ts
```diff
@@ -160,7 +160,7 @@ export class RoleService {
 
     if (!enrolmentPreconditions || enrolmentPreconditions.length < 1) return;
     for (const { type, conditions } of enrolmentPreconditions) {
-      if (type === 'role' && conditions) {
+      if (type === 'role' && conditions && conditions?.length > 0) {
```
JGiter, line 163: Is it possible that for some condition type there are no conditions? By the way don't you know why `enrollmentPreconditions` are optional? Being it required we might just iterate over them without any checking like this
`for ({type, conditions} of enrollmentPreconditions) {
   return didDocument.service.some ...
}`

jrhender: I think in the code that you suggested, this object would return `false` (indicating that the conditions aren't met), right. IMO, it should return `true` (which is the purpose of this fix) because a 'role` condition with an empty `conditions` array isn't a meaningful condition.
```
enrolmentPreconditions = {
  type: 'role',
  conditions: []
}
```

I'm open to making `IRoleDefintion.enrolmentPreconditions` mandatory, but I think we should get this fix in first and then make that change separately as I think that this bug is currently blocking enrolment requests as, currently, roleDefinitions without preconditions have the structure above

JGiter: Ok. Lets improve this latter. Also I think it should not be allowed to create preconditions without conditions

JGiter (approved):

PR #101, opened 2021-04-09
fix(assets.controller.getByID): check for null asset

jrhender:
https://sentry.io/organizations/energy-web/issues/2323842027/?project=5651522

### DIFF
--- a/src/modules/assets/assets.controller.ts
+++ b/src/modules/assets/assets.controller.ts
@@ -22,13 +22,13 @@ import { Order } from './assets.types';
 @ApiTags('Assets')
 @Controller('assets')
 export class AssetsController {
-  constructor(private readonly assetsService: AssetsService) {}
+  constructor(private readonly assetsService: AssetsService) { }
 
   @Get(':id')
   async getByID(@Param('id') id: string, @User() currentUser?: string) {
     const asset = await this.assetsService.getById(id);
     if (
-      currentUser &&
+      currentUser && asset &&
       asset.offeredTo !== currentUser &&
       asset.owner !== currentUser
     ) {


src/modules/assets/assets.controller.ts
```diff
@@ -22,13 +22,13 @@ import { Order } from './assets.types';
...
 export class AssetsController {
-  constructor(private readonly assetsService: AssetsService) {}
+  constructor(private readonly assetsService: AssetsService) { }
 
   @Get(':id')
   async getByID(@Param('id') id: string, @User() currentUser?: string) {
     const asset = await this.assetsService.getById(id);
     if (
-      currentUser &&
+      currentUser && asset &&
```
JGiter, line 31: Do I understand correctly that only owner or offeredTo should be allowed to get asset info? If so, then unauthenticated user (which doesn't provide his `did`) will pass this check

jrhender: I think that it is not possible for an unauthenticated user to call this method unless the cache server is configured to not use authentication. So I think this implementation is okay (other than the `asset == undefined` issue that this PR fixes).
My assumption on how this works is that the `currentUser` is injected by the inclusion of the `@User()` annotation. I believe `@User` itself is populated by the auth process. As this controller has the `@Auth` annotation, then `@User` should always be populated if auth is used. I believe the rational for having `currentUser` be optional is to provide the option to run the cache server without auth.

JGiter: Alright then. It seems like in theory though that any authenticated user is able to get asset, because `@User` doesn't require user did in the request even if `@Auth` is enabled

JGiter (approved):

PR #104, opened 2021-04-14
feature/asset_enrollment

JGiter:
Changes related to asset enrollment:

- To avoid spamming claim issuer cache server will verify that either claim requested by claim subject or by subject owner
- On issuing claim subject will be notified along with requester


jrhender (approved):
Looks good 👍

PR #106, opened 2021-04-16
feature/improve claim api

JGiter:
According to JWT specification token subject contained as part of token. As subject is often required part of claim it is better extract  subject and add it to other fields of `Claim` entity on load from database

### DIFF
--- a/src/modules/claim/claim.service.ts
+++ b/src/modules/claim/claim.service.ts
@@ -281,6 +281,46 @@ export class ClaimService {
   }
 
   /**
+   * Get claims issued for given subject
+   * @param did subjects's DID
+   * @param filters additional filters
+   */
+  async getBySubject({
+    subject,
+    filters = {},
+    currentUser,
+  }: {
+    subject: string;
+    filters?: QueryFilters;
+    currentUser?: string;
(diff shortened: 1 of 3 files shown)


src/modules/claim/claim.service.ts
```diff
@@ -280,6 +280,46 @@ export class ClaimService {
...
+    }
+
+    if (currentUser) {
+      qb.andWhere(new Brackets(query => {
+        query.where(':currentUser = ANY ("claimIssuer")', {
+          currentUser,
+        }).orWhere(':currentUser = requester', { currentUser })
+      }))
+    }
+    return (await qb.getMany()).filter((claim) => claim.subject === subject);
```
jrhender, line 320: IMO, I would add a comment here (maybe a TODO comment) mentioning that this should be a part of the database query (as I think you explained in the PR comment)

jrhender (approved):

src/modules/claim/claim.service.ts
JGiter: The problem is that subject doesn't stored in database. It is extracted from token, when claim loaded from db

PR #108, opened 2021-04-20
bug/asset transfer events are not published

JGiter:
Publish offered, offerAccepted, offerRejected, OfferCanceled events

### DIFF
--- a/src/modules/assets/assets.subscriber.ts
+++ b/src/modules/assets/assets.subscriber.ts
@@ -5,12 +5,15 @@ import { Repository } from 'typeorm';
 import { Logger } from '../logger/logger.service';
 import { AssetsHistory } from './assets.entity';
 import { AssetEvent, AssetHistoryEventType } from './assets.event';
+import { NatsService } from '../nats/nats.service';
+import { NATS_EXCHANGE_TOPIC } from '../claim/claim.types';
 
 @Injectable()
 export class AssetsEventSubscriber {
   constructor(
     @InjectRepository(AssetsHistory)
     private readonly historyRepository: Repository<AssetsHistory>,
+    private readonly nats: NatsService,
     private readonly logger: Logger,
   ) {
     this.logger.setContext(AssetsEventSubscriber.name);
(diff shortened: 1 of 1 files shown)


src/modules/assets/assets.subscriber.ts
```diff
@@ -130,8 +157,54 @@ export class AssetsEventSubscriber {
...
+    }
+    const eventToSave = AssetsHistory.create({
+      ...event,
+      type: AssetHistoryEventType.ASSET_OFFER_CANCELED,
+    });
+    const saved = await this.historyRepository.save(eventToSave);
+    this.logger.debug(
+      `${eventToSave.type} event for ${
+      eventToSave.assetId
+      } handled: ${JSON.stringify(saved)}`,
```
jrhender, line 195:
```suggestion
      } saved: ${JSON.stringify(saved)}`,
```
Minor point, but I would use "saved" instead of "handled" here as technically the "handled" function is executed even if `isEventAlreadySaved == true`

src/modules/assets/assets.subscriber.ts
```diff
@@ -86,9 +89,21 @@ export class AssetsEventSubscriber {
...
+      event: MessageType.AssetOffered,
+      asset: event.assetId,
+      owner: event.emittedBy,
+      offeredTo: event.relatedTo,
+      at: event.at
+    };
+    this.nats.connection.publish(
+      `${msg.offeredTo}.${NATS_EXCHANGE_TOPIC}`,
+      JSON.stringify(msg)
+    );
```
jrhender, line 106: It looks like this code is duplicated a few times. Maybe it makes sense to have function like `publishMessage(type, targetDID)`?

src/modules/assets/assets.subscriber.ts
jrhender, line 195: It looks like this bit is also duplicated and could be extracted into a method. In fact, maybe the whole `handle` methods could be a common function that just differs based on the event type. I think it could make it easier to maintain, but up to you.

jrhender (commented):
Looks pretty good so far

jrhender:
> * [x]  define enumeration to differentiate message type
> * [x]  publish offered, offerAccepted, offerRejected, OfferCanceled events

@JGiter Should this PR still be in draft? It looks like both steps are done

marcin-l-tsh (approved):

src/modules/assets/assets.subscriber.ts
JGiter: Great notice. Code reduced half-size

jrhender (approved):
Looks good to me

PR #112, opened 2021-04-26
bug/same enroll request

JGiter:
Enrollment request with same `subject`, role and role version can be send twice

### DIFF
--- a/src/modules/claim/claim.service.ts
+++ b/src/modules/claim/claim.service.ts
@@ -16,10 +16,13 @@ import { InjectQueue } from '@nestjs/bull';
 import { Queue } from 'bull';
 import { NatsService } from '../nats/nats.service';
 import jwt from 'jsonwebtoken';
+import { v5 } from 'uuid';
 interface QueryFilters {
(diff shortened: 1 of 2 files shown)


src/modules/claim/claim.service.ts
```diff
@@ -352,4 +356,10 @@ export class ClaimService {
     });
     return claims.map(claim => claim.requester);
   }
+
+  static idOfClaim(claimReq: IClaimRequest) {
+    const { token, claimType: role, claimTypeVersion: version } = claimReq;
+    const { sub: subject } = jwt.decode(token);
+    return v5(JSON.stringify({ subject, role, version }), UUID_NAMESPACE);
```
jrhender, line 363: This is deterministic then? For the the same subject, role & version, it will create the same uuid?

JGiter: > This is deterministic then? For the the same subject, role & version, it will create the same uuid?

Yes, it should prevent create same enrollment request

jrhender (approved):
Looks good

PR #117, opened 2021-05-17
fix(SWTCH-891): Change “Asset_Transferred” to “Offered From”

DawidGil-jlabs:


src/modules/assets/assets.service.ts
```diff
@@ -310,7 +310,7 @@ export class AssetsService {
         this.assetsRepository.save(update);
 
         this.eventEmitter.emit(
-          AssetHistoryEventType.ASSET_TRANSFERRED,
+          AssetHistoryEventType.OFFERED_FROM,
```
jrhender, line 313: This change doesn't really make sense to me. The smart contract event name is `'IdentityTransferred'` so don't understand why we would emit an `OFFERED_FROM` event. @DawidGil-jlabs could you please explain?

src/migrations/1617092420994-AddAssets.ts
```diff
@@ -6,7 +6,7 @@ export class AddAssets1617092420994 implements MigrationInterface {
...
         await queryRunner.query(`CREATE TABLE "did_document_entity" ("id" character varying NOT NULL, "service" jsonb NOT NULL, "authentication" jsonb NOT NULL, "created" character varying, "delegates" text array, "proof" jsonb, "publicKey" jsonb NOT NULL, "updated" character varying, "@context" character varying NOT NULL, "logs" character varying NOT NULL, CONSTRAINT "PK_d96048f4c93be203eeff05ef404" PRIMARY KEY ("id"))`);
         await queryRunner.query(`CREATE TABLE "asset" ("id" text NOT NULL, "owner" character varying NOT NULL, "offeredTo" character varying, "createdAt" character varying NOT NULL, "updatedAt" character varying NOT NULL, "documentId" character varying, CONSTRAINT "REL_c82497e54ba94a090f9c5f4bff" UNIQUE ("documentId"), CONSTRAINT "PK_1209d107fe21482beaea51b745e" PRIMARY KEY ("id"))`);
-        await queryRunner.query(`CREATE TYPE "assets_history_type_enum" AS ENUM('ASSET_CREATED', 'ASSET_OFFERED', 'ASSET_OFFER_CANCELED', 'ASSET_TRANSFERRED', 'ASSET_OFFER_REJECTED')`);
+        await queryRunner.query(`CREATE TYPE "assets_history_type_enum" AS ENUM('ASSET_CREATED', 'ASSET_OFFERED', 'ASSET_OFFER_CANCELED', 'OFFERED_FROM', 'ASSET_OFFER_REJECTED')`);
```
jrhender, line 9: I think would make sense to generate a new migration for change like this instead of editing an existing migration

src/graphql/schema.gql
```diff
@@ -198,7 +198,7 @@ enum AssetHistoryEventType {
   ASSET_CREATED
   ASSET_OFFERED
   ASSET_OFFER_CANCELED
-  ASSET_TRANSFERRED
+  OFFERED_FROM
```
jrhender, line 201: Personally, I like the `ASSET_` prefix if it is an event related to assets

jrhender (commented):
@DawidGil-jlabs Nice first `iam-cache-server` PR 👍 . A few comments to address.

PR #118, opened 2021-05-18
feature/get all enrollments

JGiter:
Existing claims endpoints doesn't allow to get enrollments for group of assets in one request. It is needed in particular to get claims of offered or previously owned assets. Its seems simpler then adding query parameters to `asset` endpoint especially if in future assets will need to be filtered by other criteria

### DIFF
--- a/src/modules/claim/claim.service.ts
+++ b/src/modules/claim/claim.service.ts
@@ -140,6 +140,13 @@ export class ClaimService {
     const updatedClaim = Claim.create({ ...claim, ...data, isAccepted: true });
     return this.claimRepository.save(updatedClaim);
   }
+  
+  /**
+   * returns all claims
+   */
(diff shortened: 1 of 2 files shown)


src/modules/claim/claim.service.ts
```diff
@@ -140,6 +140,13 @@ export class ClaimService {
     const updatedClaim = Claim.create({ ...claim, ...data, isAccepted: true });
     return this.claimRepository.save(updatedClaim);
   }
+  
+  /**
+   * returns all claims
```
jrhender, line 145: @JGiter Could you clarify what you mean by `all claims` here? Is this "all claims for a specific DID" or "maybe all claims for the currently signed-in DID"?

JGiter: It returns all claims of all DID's. I've updated comment and PR description

PR #121, opened 2021-05-21
bug/inconsistent claim filter flags

JGiter:
https://energyweb.atlassian.net/browse/SWTCH-841

### DIFF
--- a/src/modules/claim/claim.controller.ts
+++ b/src/modules/claim/claim.controller.ts
@@ -228,16 +228,15 @@ export class ClaimController {
   })
   public async getByIssuerDid(
     @Param('did') issuer: string,
-    @Query('accepted', BooleanPipe)
-    accepted?: boolean,
-    @Query('namespace') parentNamespace?: string,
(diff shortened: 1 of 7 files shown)


src/modules/claim/claim.controller.ts
```diff
@@ -249,27 +248,26 @@ export class ClaimController {
...
   @ApiQuery({
-    name: 'accepted',
+    name: 'isAccepted',
     required: false,
     description:
       '**true** - show only accepted <br> **false** - show only pending',
   })
   @ApiQuery({
-    name: 'namespace',
+    name: 'parentNamespace',
```
jrhender, line 257: @JGiter Does the use of `parentNamespace` make sense here? I think that `namespace` is actually more clear. In what sense is it the `parent`?

JGiter: You are right. This is just app or org where corresponding role is defined

jrhender (approved):
Looks good to me

PR #122, opened 2021-05-26
feature/SWTCH-932 claim subject column

JGiter:
Adds `subject` column in `claims` table and populates it with `sub` of decoded claim's token

Blocks: https://github.com/energywebfoundation/iam-cache-server/pull/120

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,8 +32,9 @@
     "docker:down": "docker-compose down -v",
     "build:docs": "rimraf docs/api && typedoc --plugin typedoc-plugin-markdown --hideBreadcrumbs true",
     "generate:jwtkeys": "openssl genrsa -out private.pem 2048 && openssl rsa -in private.pem -pubout > public.pem",
+    "migrate:claim-subject": "node src/scripts/migrate_claim_subject.js",
     "migration:generate": "docker-compose run --rm app ./node_modules/.bin/typeorm migration:generate -n",
-    "migration:run": "docker-compose run --rm app ./node_modules/.bin/typeorm migration:run",
+    "migration:run": "docker-compose run --rm app ./node_modules/.bin/typeorm migration:run && npm run migrate:claim-subject",
     "migration:revert": "docker-compose run --rm app ./node_modules/.bin/typeorm migration:revert",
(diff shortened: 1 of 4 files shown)


src/scripts/add_claim_subject_column.js
```diff
@@ -0,0 +1,38 @@
...
+    "entities": [
+      "dist/**/*.entity.js"
+    ]
+  });
+  const claimsRepository = connection.getRepository(Claim);
+  let claims = await claimsRepository.find();
+
+  const queryRunner = connection.createQueryRunner();
+  if (!(await queryRunner.hasColumn("claim", "subject"))) {
+    await queryRunner.addColumn("claim", new TableColumn({ name: 'subject', type: 'varchar', default: "''" }));
```
jrhender, line 24: Why adding the column via this script instead of typeorm migration? IMO, my first impression is that a typeorm migration would be better so that schema changes are in one place.

src/scripts/add_claim_subject_column.js
```diff
@@ -0,0 +1,38 @@
...
+  }
+  
+  const jwt = new JWT(new Keys());
+  for await (const claim of claims) {
+    const payload = jwt.decode(claim.token);
+    let subject = payload.sub;
+    if (!subject || subject.length === 0 || !subject.startsWith("did")) {
+      subject = payload.iss;
+      await claimsRepository.update(claim.id, { ...claim, subject });
```
jrhender, line 33: Shouldn't this be outside of the `if` branch? For this script, we always want to update the claim `subject` as it wasn't set at all before?

jrhender (commented):
Were you able to find if typeorm has some support for scripts? From what I can tell they don't.

JGiter:
> Were you able to find if typeorm has some support for scripts? From what I can tell they don't.

As I understand all typeorm needs is properly configured connection and entity schemes. I tried this script for and it works as expected

src/scripts/add_claim_subject_column.js
JGiter: My bad

jrhender (approved):
Looks good

PR #123, opened 2021-05-27
feature/SWTCH-965 deploy claim manager

JGiter:


jrhender (approved):
I guess you're thinking that we will eventually need to cache roles from the `ClaimManager`?

JGiter:
> I guess you're thinking that we will eventually need to cache roles from the `ClaimManager`?

Yes. Also it can be used to notify requester about registered role

PR #126, opened 2021-06-01
feature/SWTCH-906 register on-chain claim request

JGiter:
- [x] add registration types in Claim 
- [x] persist agreement and proof tokens and registration types

### DIFF
--- a/src/migrations/1622548828009-AddClaimRegistrationTypes.ts
+++ b/src/migrations/1622548828009-AddClaimRegistrationTypes.ts
@@ -1,0 +1,16 @@
+import {MigrationInterface, QueryRunner} from "typeorm";
+
+export class AddClaimRegistrationTypes1622548828009 implements MigrationInterface {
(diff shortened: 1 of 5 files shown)


jrhender (approved):
Looks good.
The only thought I had was whether or not we should be more specific with the types. Like maybe `OffChainJwtV1` and `OnChainV1`? Might make it easier to deferentiate different claim formats in the future. What do you think

JGiter:
> Looks good.
> The only thought I had was whether or not we should be more specific with the types. Like maybe `OffChainJwtV1` and `OnChainV1`? Might make it easier to deferentiate different claim formats in the future. What do you think

I suggest `offChainProof` and `onChainProof` as `jwt` is more about implementation details then purpose. And I would avoid such complication as versioning of registration types. If registration process changes then its better to migrate previously registered roles

ahmedolaibrahim (approved):

PR #127, opened 2021-06-01
fix(ens.service): fix import of DomainNotifier from iam-contracts

jrhender:
SWTCH-721


jrhender:
Merging as is minor change and near end of sprint

PR #128, opened 2021-06-03
docs: update dev setup in readme

jrhender:
WIP to improve iam-cache-server dev instructions. Let me know what you think

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,21 +7,25 @@
 # development
 $ git checkout develop
 $ cp .env.dist .env
-$ cp docker-compose.dev.yml docker.compose.yml
+$ cp docker-compose.dev.yml docker-compose.yml
 ```
 
 ```bash
 # production mode
 $ git checkout master
 $ cp .env.dist .env
-$ cp docker-compose.prod.yml docker.compose.yml
+$ cp docker-compose.prod.yml docker-compose.yml
 ```
 
 ### Populate env file for development
 
 The following values can be used in the `.env` file
+
+#### NESTJS CONFIG
+NESTJS_PORT=
+
 #### DB Config
-DB_HOST=localhost
+DB_HOST=postgres
 DB_PORT=5432
 DB_USERNAME=postgres
 DB_PASSWORD=password
@@ -31,7 +35,12 @@ DB_NAME=dev
 REDIS_VERSION=6.0.8 
 REDIS_PASSWORD="password"
 REDIS_PORT=6379
-REDIS_HOST=localhost
+REDIS_HOST=redis
+
+#### NATS.IO CONFIG
+NATS_VERSION=2.1.8
+NATS_CLIENTS_URL=nats:4222
+
 
 ## Installation
 
@@ -42,6 +51,10 @@ $ npm run generate:jwtkeys
 ```
 
 ## Running the app
+
+### Populate env file for development
+
+The values from .env.dev can be used to run locally
 
 ```bash
 # development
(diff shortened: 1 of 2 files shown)


knzeng-e (commented):
I do appreciate the separation of .env files. It brings more precisions without overloading the main doc

README.md
```diff
@@ -5,16 +5,34 @@
 
 ```bash
 # development
+$ git checkout develop
 $ cp .env.dist .env
 $ cp docker-compose.dev.yml docker.compose.yml
```
knzeng-e, line 10: should'nt it be `cp docker-compose.dev.yml docker-compose.yml` here ? ( '-' instead of '.')

.env.dev
```diff
@@ -0,0 +1,17 @@
+# DB CONFIG
```
knzeng-e, line 1: Should we also provide `DGRAPH_GRPC_HOST` & `NESTJS_PORT` in this file ?

# NESTJS_CONFIG
NESTJS_PORT=...

# DGRAPH CONFIG
DGRAPH_GRPC_HOST=...

README.md
jrhender: yes, will make this change, thanks

docker/dev/Dockerfile
```diff
@@ -2,8 +2,7 @@ FROM node:12.13-alpine As development
 
 WORKDIR /app
 
-RUN apk add --no-cache bash curl openssl python make g++ && \
-  curl -sfL https://install.goreleaser.com/github.com/tj/node-prune.sh | sh && \
+RUN apk add --no-cache bash curl openssl python make g++ && npm prune && \
```
ahmedolaibrahim, line 5: I had to include this because this package `https://install.goreleaser.com/github.com/tj/node-prune.sh` does not support linux arm64 devices which is the architecture of apple m1.  using npm prune which basically does the same thing fixes this.

jrhender:
@ahmedolaibrahim It seems like `node-prune.sh` removes files from the node_modules (https://github.com/tj/node-prune#what) whereas `npm prune` is removing the extraneous packages themselves (https://docs.npmjs.com/cli/v7/commands/npm-prune). So they seem to be slightly different. Is that how you see it as well? Or am I misunderstanding?

.env.dev
ahmedolaibrahim: I think `NESTJS_PORT` is needed but not the DGRAPH config values as I was able to set up the docker env without it.

ahmedolaibrahim:
> > https://docs.npmjs.com/cli/v7/commands/npm-prune
> 
> > @ahmedolaibrahim It seems like `node-prune.sh` removes files from the node_modules (https://github.com/tj/node-prune#what) whereas `npm prune` is removing the extraneous packages themselves (https://docs.npmjs.com/cli/v7/commands/npm-prune). So they seem to be slightly different. Is that how you see it as well? Or am I misunderstanding?
> 
> @jrhender My thinking is `npm prune` is similar to `node-prune` in the sense that it removes unwanted packages prior to `npm install ` command which eventually generates the `node_modules` folder thereby preventing unwanted files. Also `npm prune` was listed as an alternative to yarn `autoclean` which is an alternative to `node-prune` lol. will have to do some more research to confirm that this is absolutely the case.

@jrhender  I have now updated the PR to now use a very very similar library that supports linux/arm64 architecture. LOL

ahmedolaibrahim (approved):

PR #131, opened 2021-06-07
Bug/swtch 982/fix enrolment

jrhender:
iam-client-lib is submitting versions as numbers so convert to string until we migrate `claimTypeVersion` column.
Also disabled failing tests so that `npm run test` passes (still need to run on ci build, should be done by https://energyweb.atlassian.net/browse/SWTCH-840 )

### DIFF
--- a/src/modules/claim/claim.dto.ts
+++ b/src/modules/claim/claim.dto.ts
@@ -9,6 +9,7 @@ import { ApiProperty } from '@nestjs/swagger';
 
 export class ClaimRequestDTO implements IClaimRequest {
   static async create(data: Partial<ClaimRequestDTO>) {
+    data.claimTypeVersion = data.claimTypeVersion.toString();
     const dto = new ClaimRequestDTO();
     Object.assign(dto, data);
     await validateOrReject(dto, { whitelist: true });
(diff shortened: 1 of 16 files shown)


src/modules/claim/claim.dto.ts
```diff
@@ -2,13 +2,15 @@ import { IClaimIssuance, IClaimRejection, IClaimRequest } from './claim.types';
...
   IsBoolean,
+  IsNumberString,
   IsString,
   validateOrReject,
 } from 'class-validator';
 import { ApiProperty } from '@nestjs/swagger';
 
 export class ClaimRequestDTO implements IClaimRequest {
   static async create(data: Partial<ClaimRequestDTO>) {
+    data.claimTypeVersion = data.claimTypeVersion.toString().split('.')[0];
```
JGiter, line 13: Wouldn't this lead to duplicate enrollment requests? If versions in the role definition and in the claim are different, how  to determine that enrollment request already submitted?
Also `claimTypeVersion` in iam-client-lib just defined as number, but it sends to cache-server whatever it receives from switchboard-dapp

jrhender: I don't think this really impacts duplicate enrolment requests as long as the check on the frontend (which is the only place that we have the check currently) is correct. I'm making a change to that here: https://github.com/energywebfoundation/switchboard-dapp/pull/293

JGiter (approved):

PR #132, opened 2021-06-08
feature/SWTCH-907 claims by issuer from role def

JGiter:
https://energyweb.atlassian.net/browse/SWTCH-907

### DIFF
--- a/src/modules/claim/claim.service.ts
+++ b/src/modules/claim/claim.service.ts
@@ -366,11 +366,11 @@ export class ClaimService {
   private async rolesByIssuer(issuer: string, namespace?: string): Promise<Role[]> {
     const rolesOfIssuer = (await this.getBySubject({ subject: issuer, filters: { isAccepted: true } }))
       .map((r) => r.claimType);
-    return (await this.roleService.getAll())
-      .filter((r: Role) => !namespace || r.namespace === namespace)
+    const roles = (await this.roleService.getAll())
       .filter(
         (r) => r.definition.issuer.did?.includes(issuer) || rolesOfIssuer.includes(r.definition.issuer.roleName)
       );
+    return namespace ? roles.filter((r: Role) => r.namespace === namespace) : roles;
   }
 
   private async filterUserRelatedClaims(currentUser: string, qb: SelectQueryBuilder<Claim>) {
--- a/src/modules/claim/claim.service.ts
+++ b/src/modules/claim/claim.service.ts
@@ -11,13 +11,14 @@ import { Logger } from '../logger/logger.service';
 import { ClaimIssueDTO, ClaimRejectionDTO, ClaimRequestDTO } from './claim.dto';
 import { Claim } from './claim.entity';
 import { InjectRepository } from '@nestjs/typeorm';
-import { Brackets, Repository } from 'typeorm';
+import { Brackets, Repository, SelectQueryBuilder } from 'typeorm';
 import { InjectQueue } from '@nestjs/bull';
 import { Queue } from 'bull';
 import { NatsService } from '../nats/nats.service';
 import jwt from 'jsonwebtoken';
 import { v5 } from 'uuid';
 import { AssetsService } from '../assets/assets.service';
+import { Role } from '../role/role.entity';
 interface QueryFilters {
   isAccepted?: boolean;
   namespace?: string;
@@ -160,22 +161,7 @@ export class ClaimService {
     qb.where('claim.subject IN (:...subjects)', { subjects });
 
     if (currentUser) {
-      const ownedAssets = (await this.assetService.getByOwner(currentUser)).map((a) => a.id);
-      const offeredAssets = (await this.assetService.getByOfferedTo(currentUser)).map((a) => a.id);
-
-      qb.andWhere(new Brackets(query => {
-        query.where(':currentUser = ANY (claim.claimIssuer)', {
-          currentUser,
-        })
-          .orWhere('claim.subject = :currentUser', { currentUser })
-          .orWhere('claim.requester = :currentUser ', { currentUser });
-        if (ownedAssets.length > 0) {
-          query.orWhere('claim.subject IN (:...ownedAssets)', { ownedAssets });
-        }
-        if (offeredAssets.length > 0) {
-          query.orWhere('claim.subject IN (:...offeredAssets)', { offeredAssets });
-        }
-      }));
+      await this.filterUserRelatedClaims(currentUser, qb);
     }
 
     if (isAccepted !== undefined) {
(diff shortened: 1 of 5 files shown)


src/modules/claim/claim.service.ts
```diff
@@ -258,28 +243,35 @@ export class ClaimService {
...
     if (isAccepted !== undefined) {
-      qb.andWhere('"isAccepted" = :isAccepted', {
+      qb.andWhere('claim.isAccepted = :isAccepted', {
         isAccepted,
       });
     }
 
     if (namespace) {
-      qb.andWhere('"namespace" = :namespace', {
+      qb.andWhere('claim.namespace = :namespace', {
```
jrhender, line 268: Isn't of filtering here, maybe could filter at the point of `roleService.getAll()`. Seems a little silly to process all the roles looking for ones this DID can issue when we only care about one namespace. Would be faster to only check if the DID can issue that namespace/role.

src/modules/claim/claim.service.ts
```diff
@@ -379,4 +371,23 @@ export class ClaimService {
...
+      })
+        .orWhere('claim.subject = :currentUser', { currentUser })
+        .orWhere('claim.requester = :currentUser ', { currentUser });
+      if (ownedAssets.length > 0) {
+        query.orWhere('claim.subject IN (:...ownedAssets)', { ownedAssets });
+      }
+      if (offeredAssets.length > 0) {
+        query.orWhere('claim.subject IN (:...offeredAssets)', { offeredAssets });
+      }
+    }));
```
jrhender, line 391: I think this might not work for this new approach. I mean, if we are not adding the `claim.claimIssuer` ahead of time, wouldn't none of these case be met when we are trying to dynamically query the issuer?

jrhender (commented):
A few comments but looks like a good start

src/modules/claim/claim.service.ts
JGiter: I think that this flag is intended to filter roles allowed to issue, not roles of issuer

src/modules/claim/claim.service.ts
JGiter: As I understand, only first clause of query verifies claim issuer, while others doesn't depend on result of this verification

src/modules/role/role.dto.ts
```diff
@@ -99,7 +99,8 @@ export class IssuerDTO implements Issuer {
   issuerType: string;
 
   @IsArray()
-  did: string[];
+  @IsOptional()
+  did?: string[];
```
jrhender, line 103: If my above comment is correct, do we even need the `did` property at all any more?
```suggestion
```

src/modules/claim/claim.service.ts
```diff
@@ -379,4 +361,36 @@ export class ClaimService {
...
+    const ownedAssets = (await this.assetService.getByOwner(currentUser)).map((a) => a.id);
+    const offeredAssets = (await this.assetService.getByOfferedTo(currentUser)).map((a) => a.id);
+
+    const rolesByUser = (await this.rolesByIssuer(currentUser)).map((r) => r.namespace);
+
+    qb.andWhere(new Brackets(query => {
+      query.where(':currentUser = ANY (claim.claimIssuer)', {
+        currentUser,
+      })
+        .orWhere('claim.namespace IN (:...rolesByUser)', { rolesByUser })
```
jrhender, line 385: Looking again at `getBySubject()`, I think it will retrieve all of the approved claims that cache-server knows about for both off-chain and on-chain roles, right? So I don't think we even need to check the `claimIssuer` property anymore, right?
```suggestion
      query.where('claim.namespace IN (:...rolesByUser)', { rolesByUser })
```

jrhender (commented):
Looks like a step in the right direction. I'm wondering if we can go even one step further and remove the use of `did` and `claimIssuer` from cache-server

src/modules/role/role.dto.ts
JGiter: I agree. Do we also need to update the `IssuerResolver.sol` or, for now, is it enough to remove the option of setting list of issuers on the front-end?

src/modules/claim/claim.service.ts
JGiter: Hmm. I am not sure. Having `claim.namespace` in `rolesByIssuer` depends on role definitions and claims of current user, while presence of current user in `claim.claimIssuer` arbitrarily set by requester

src/modules/role/role.dto.ts
jrhender: Ah sorry, I think I was confused. This `did` property is on the role definition which hasn't changed (role definitions can indeed have a list of DIDs as the valid issuers. So I think we still need the `did` property.

I guess you're making it optional because role can either have a `did` array or a `roleName` as the issuer?

jrhender (approved):
Still have my outstanding comment about not checkin the `claimIssuer` in `claim.service` but I think that it an optional improvement. Looks good to me. 👍

src/modules/role/role.dto.ts
JGiter: Yes, I want role definition to conform to `IRoleDefinition`. May its better to import this interface from iam-contract. In order to do this though `enrollmentPrecondition`, `metadata` and `fields` are needs to be migrated as they are optional in cache-server. Better to do this in separate PR

src/modules/claim/claim.service.ts
JGiter: You are right, `claim.claimIssuer` filled from role issuers list. Since this property is of no use, maybe remove it?

src/modules/role/role.dto.ts
jrhender: Ok, sounds good

src/modules/claim/claim.service.ts
```diff
@@ -379,4 +362,33 @@ export class ClaimService {
...
+  private async filterUserRelatedClaims(currentUser: string, qb: SelectQueryBuilder<Claim>) {
+    const ownedAssets = (await this.assetService.getByOwner(currentUser)).map((a) => a.id);
+    const offeredAssets = (await this.assetService.getByOfferedTo(currentUser)).map((a) => a.id);
+
+    const rolesByUser = (await this.rolesByIssuer(currentUser)).map((r) => r.namespace);
+
+    qb.andWhere(new Brackets(query => {
+      query.where('claim.claimType IN (:...rolesByUser)', { rolesByUser })
+        .orWhere('claim.subject = :currentUser', { currentUser })
+        .orWhere('claim.requester = :currentUser ', { currentUser });
```
jrhender, line 385: Possibly a minor point, but maybe remove the space here. Not sure if it would cause an issue or not.
```suggestion
        .orWhere('claim.requester = :currentUser', { currentUser });
```

jrhender (approved):
Looks good

PR #135, opened 2021-06-14
feature/SWTCH-840_testing_framework

knzeng-e:
iam-cache-server has no automated tests setup. This makes is more difficult test new functionality as well as guard against regressions.

Acceptance Criteria:

* Automated test harness which executes using actual Postgres and NATS

* Tests are run during CI builds (using npm test)

* Only a small number of tests are required. This is more about the test harness setup.


docker-compose.dev.yml
```diff
@@ -10,29 +10,60 @@ services:
...
-      - .env
+      - .env.ci
     networks:
       - main
     depends_on:
+      - dgraph
+      - zero
       - nats
-      - redis
+      - ratel
```
jrhender, line 24: Why the addition of the dgraph components? I think they are not needed anymore 😃

.github/workflows/cache_server.test.yml
```diff
@@ -0,0 +1,66 @@
...
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '14'
+
+      # Runs a single command using the runners shell
+      - name: Install dependencies
+        run: npm install
+        
+      - name: Docker Build
+        run: npm run docker:build -f "docker-compose.dev.yml"
```
jrhender, line 53: I'm a bit confused. We're running the whole dev `docker-compose` but we're also running services? I guess what would make sense to me is if we run the `docker-compose.prod.yml` (as that is what we are running in production) and then only run `postgres` as a service in the GHA. How does that sound?

jrhender (commented):
Looks like a good start! I think the next thing to do would be to add an `e2e` test to see it working. It looks like Origin has some good examples here: https://github.com/energywebfoundation/origin/tree/master/packages/origin-backend/test

.github/workflows/cache_server.test.yml
knzeng-e: Yes that sounds indeed balanced. Is it ok if I provide a docker-compose.ci.yml ? Since I didn't use required values like AWS ECR credentials.

.github/workflows/cache_server.test.yml
```diff
@@ -0,0 +1,49 @@
...
+      - uses: actions/checkout@v2
+
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '14'
+
+      - name: Install dependencies
+        run: npm install
+        
+      - name: Docker Build
```
ahmedolaibrahim, line 36: i think we shouldn't use docker as part of this workflow because it will increase the time it takes for this workflow to build. let the self hosted containers by the actions handle that.

ahmedolaibrahim (commented):
Looking good just a suggestion around using docker as part of the actions workflow.  Since the end goal of this is tests, we should leave that to the workflow action and use services to add dependencies to the workflow. Adding docker to the workflow will increase the build time and it's use should be limited to the dev and main deployment process.

.github/workflows/cache_server.test.yml
```diff
@@ -0,0 +1,51 @@
...
+       env:
+        DB_USERNAME: ${{secrets.DB_USERNAME}}
+        DB_PASSWORD: ${{secrets.DB_PASSWORD}}
+        DB_HOST: 'postgres'  
+       ports:
+       - 5432:5432
+      
+    runs-on: ubuntu-latest
+    steps:
+      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
```
JGiter, line 26: I guess this comment isn't needed

docker-compose.dev.yml
```diff
@@ -10,29 +10,60 @@ services:
       - .:/app
       - /app/node_modules
     ports:
-      - ${NESTJS_PORT}:${NESTJS_PORT}
```
JGiter, line 13: why not use environment variables?

docker-compose.dev.yml
```diff
@@ -10,29 +10,60 @@ services:
...
   postgres:
-    container_name: postgres
     image: postgres:12-alpine
     environment:
-      POSTGRES_PASSWORD: ${DB_PASSWORD}
-      POSTGRES_USERNAME: ${DB_USERNAME}
-      POSTGRES_DB: ${DB_NAME}
-      POSTGRES_PORT: ${DB_PORT}
+      #POSTGRES_PASSWORD: secrets.DB_PASSWORD
+      # POSTGRES_USERNAME: secrets.DB_USERNAME
```
JGiter, line 30: unnecessary comments

docker-compose.dev.yml
```diff
@@ -38,13 +38,13 @@ services:
...
+    command: '-c /etc/nats/nats.conf'
     networks:
       - main
     ports:
-      - "4222:4222"
-      - "8222:8222"
-      - "9222:9222"
+      - '4222:4222'
+      - '8222:8222'
+      - '9222:9222'
```
hrachocode, line 47: I think we should all have the same autoformatting configs so all code styling will be similar, for example now in my case autosave replaces double quotes with single quotes, and yours doing the opposite.

ahmedolaibrahim: yeah I agree with you but in this case the prettier config [rule](https://github.com/energywebfoundation/iam-cache-server/blob/6ed1cea58471fa4c6a765d13bfba94c8a9c84c53/.prettierrc#L2) in this repo is set to single quotes. i'm guessing the files were created before the rules were added.

hrachocode: ok thank you.

src/modules/organization/organization.controller.e2e.spec.ts
```diff
@@ -0,0 +1,164 @@
...
+    module = await Test.createTestingModule({
+      imports: [
+        TypeOrmModule.forRoot(TestDbCOnfig.default as TypeOrmModuleOptions),
+        TypeOrmModule.forFeature([Organization, Application]),
+        LoggerModule,
+        ConfigModule,
+        SentryModule,
+      ],
+      controllers: [OrganizationController],
+      providers: [ConfigService, OrganizationService],
```
JGiter, line 65: isn't ConfigService exported from ConfigModule?

.github/workflows/cache_server.test.yml
```diff
@@ -0,0 +1,56 @@
...
+        image: postgres:12
+        env:
+          POSTGRES_USER: ${{secrets.DB_USERNAME}}
+          POSTGRES_DB: dev-test
+          POSTGRES_PASSWORD: ${{secrets.DB_PASSWORD}}
+        ports:
+          - 5432:5432
+        # Add a health check
+        options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
+
```
JGiter, line 24: We will also need NATS to test claims

JGiter (changes requested):

src/modules/organization/organization.controller.e2e.spec.ts
ahmedolaibrahim: nice catch! my bad. I used it for testing something and forgot to remove it.

.github/workflows/cache_server.test.yml
ahmedolaibrahim: okay.

src/modules/organization/organization.controller.e2e.spec.ts
```diff
@@ -0,0 +1,164 @@
...
+  let organizations: Organization[];
+
+  beforeEach(async () => {
+    module = await Test.createTestingModule({
+      imports: [
+        TypeOrmModule.forRoot(TestDbCOnfig.default as TypeOrmModuleOptions),
+        TypeOrmModule.forFeature([Organization, Application]),
+        LoggerModule,
+        ConfigModule,
+        SentryModule,
```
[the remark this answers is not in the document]
ahmedolaibrahim: the `SentryModule` needs to be injected here because of the way dependency injection works. The `OrganizationModule` imports `OrganizationService` which then imports `LoggerService` which then goes on to use `SentryErrorInterceptor` which would not be available and would throw an error if the `SentryModule` is not imported in the current `TestingModule` context so it might not be needed in the other e2e tests depending or wether they use  the LoggerService or not.

JGiter (approved):

PR #136, opened 2021-06-16
Fix/swtch 1041/parallel read

jrhender:
A couple of changes both aiming to fix case where multiple did documents are read in parallel:
- [x] fix(did.service): await resolver.readFromBlock in getAllLogs
- [x] fix(did.service): update ew-did-registry packages to fix parallel read of registry

Bug: https://energyweb.atlassian.net/browse/SWTCH-967


knzeng-e (commented):
Everything looks good to me !

PR #137, opened 2021-06-17
SWTCH-1043 only return IDIDDocument

jrhender:
https://energyweb.atlassian.net/browse/SWTCH-1043

- [x] test(did.service): service defined with mock dependencies
- [x] only return IDIDDocument from getDocument

### DIFF
--- a/src/modules/did/did.service.spec.ts
+++ b/src/modules/did/did.service.spec.ts
@@ -1,13 +1,76 @@
+import { getQueueToken } from '@nestjs/bull';
+import { HttpService, Logger } from '@nestjs/common';
+import { ConfigService } from '@nestjs/config';
+import { SchedulerRegistry } from '@nestjs/schedule';
 import { Test, TestingModule } from '@nestjs/testing';
+import { getRepositoryToken } from '@nestjs/typeorm';
+import { Provider } from '../../common/provider';
+import { DIDDocumentEntity } from './did.entity';
 import { DIDService } from './did.service';
 
-// TODO: fix test so pending can be removed
-xdescribe('DidDocumentService', () => {
+const MockLogger = {
+  log: jest.fn(),
+  setContext: jest.fn(),
+};
+const MockObject = {};
+const MockConfigService = {
+  get: jest.fn((key: string) => {
+    if (key === 'DID_SYNC_ENABLED') {
+      return 'false';
+    }
+    return null;
+  }),
+};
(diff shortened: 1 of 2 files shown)


ahmedolaibrahim (commented):
The test setup looks good to me but I can't seem to find any test cases for the PR title or in the test file. is this deliberate? 
never mind, I see the marked bullet point describes what the PR does.

src/modules/did/did.service.spec.ts
```diff
@@ -1,19 +1,109 @@
...
+    const didDoc = await service.getById("<any DID>");
+    expect(didDoc).toBeDefined();
+    expect(didDoc[nameof<DIDDocumentEntity>("logs")]).toBeUndefined();
+  });
+
+  it('getByID should return only IDIDDocument params from non-cached document', async () => {
+    didDocEntity = undefined;
+    const didDoc = await service.getById("<any DID>");
+    expect(didDoc).toBeDefined();
+    expect(didDoc[nameof<DIDDocumentEntity>("logs")]).toBeUndefined();
```
ahmedolaibrahim, line 107: IMO  methods like `toBeDefined`, `toBeUndefined` don't help that much when it comes to ensuring deterministic tests. I think we should also assert for the actual values we expect and probably mock methods that get called. this would help catching type  related bugs .

src/modules/did/did.service.ts
```diff
@@ -90,13 +90,18 @@ export class DIDService {
    */
   public async getById(did: string): Promise<IDIDDocument> {
     const cachedDIDDocument = await this.didRepository.findOne(did);
-    if (cachedDIDDocument) return cachedDIDDocument;
+    if (cachedDIDDocument) {
+      delete cachedDIDDocument.logs;
```
ahmedolaibrahim, line 94: rather the returning the entire object and deleting properties we don't need, I think we could just select the fields that we need and return that.

ahmedolaibrahim (commented):
looks good. just a few minor comments

src/modules/did/did.service.ts
jrhender: @ahmedolaibrahim good point. In that way, we can be sure that we are constructing the return object that we want

src/modules/did/did.service.spec.ts
jrhender: Thanks. Refactored test to check for each property

ahmedolaibrahim (approved):

PR #138, opened 2021-06-21
fix(claim.dto): add subjectAgreement+registrationTypes

jrhender:
SWTCH-1061
Adding subjectAgreement to claim.dto so that it gets persisted to the entity


JGiter (approved):

ahmedolaibrahim (approved):

jrhender:
@JGiter `registrationTypes` were also missing from the dto so I added another commit. Could you take another look when you have a moment?

JGiter (approved):

PR #142, opened 2021-06-25
Fix/swtch 1042 include claims param

jrhender:
- fix(did.controller): removed `includeClaims` param as not respected. DID document services are being persisted directly with claimData so hard to separate at this point. Should not break anyone who is using `includeClaims` as they will still get the data
- docs(did.service): corrected some documentation

### DIFF
--- a/src/modules/did/did.service.ts
+++ b/src/modules/did/did.service.ts
@@ -83,10 +83,9 @@ export class DIDService {
   }
 
   /**
-   * Retrieves a DID Document for a given DID
-   * @param {DID} did DID whose document should be retrieved
-   * @param {boolean} enhanceWithClaims
-   * @returns {IDIDDocument} Resolved DID Document. Null if no Document is not cached.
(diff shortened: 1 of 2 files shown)


src/modules/did/did.service.ts
```diff
@@ -83,10 +83,9 @@ export class DIDService {
   }
 
   /**
-   * Retrieves a DID Document for a given DID
-   * @param {DID} did DID whose document should be retrieved
-   * @param {boolean} enhanceWithClaims
-   * @returns {IDIDDocument} Resolved DID Document. Null if no Document is not cached.
+   * Retrieves a DID Document for a given DID. Retrieves from blockchain if not chached.
```
knzeng-e, line 86: A minor typo here: Retrieves from blockchain if not `chached`. >> Retrieves from blockchain if not cached.

knzeng-e (approved):
Except the small typo in doc, it looks good to me

ahmedolaibrahim (approved):

PR #144, opened 2021-07-01
bug/SWTCH-1089 owner cant enroll asset

JGiter:
- updated claim and identity managers
- `ClaimIssuerDTO.issuedToken` set to optional. Being required it causes error when only on-chain registration requested


jrhender (approved):

src/modules/claim/claim.dto.ts
```diff
@@ -84,8 +84,9 @@ export class ClaimIssueDTO implements IClaimIssuance {
   claimIssuer: string[];
 
   @IsString()
+  @IsOptional()
```
jrhender, line 87:
```suggestion
  // Is not set in the event that only on-chain role is issued
  @IsOptional()
```

jrhender (approved):

PR #145, opened 2021-07-05
csrf protection based on origin check, initial commit

hrachocode:


src/modules/auth/token.service.ts
```diff
@@ -64,4 +69,28 @@ export class TokenService {
...
+    if (req.headers['authorization']) {
+      token = req.headers['authorization'].replace('Bearer ', '');
+    } else {
+      token = req.cookies.token
+    }
+
+    if (token) {
+      const decodedToken = this.jwtService.decode(token) as TokenPayload;
+      if (decodedToken.supportedOrigins.includes(requestOrigin) ||
+        decodedToken.supportedHosts.includes(req.hostname) // swagger and server based requests check only
```
hrachocode, line 86: the second check is written only for swagger and server-based requests

as their requests are from the same domain as the servers domain, **origin** header is missing in requests, so I am checking if the requested URL is the same as servers HOST(like server is requesting to server), then check should pass.

src/modules/auth/login.controller.ts
```diff
@@ -105,7 +106,7 @@ export class LoginController {
     );
 
     const [token, refreshToken] = await Promise.all([
-      this.tokenService.generateAccessToken({ did: userDid, verifiedRoles }),
+      this.tokenService.generateAccessToken({ did: userDid, verifiedRoles, supportedOrigins, supportedHosts }),
```
jrhender, line 109: We don't want to restrict clients to specific "supportedOrigins". I think instead the approach that we should take is that we store the origin that the client used during login and check if that is the origin of subsequent requests.
I think this is roughly what we need:
```suggestion
      const origin = req.headers['origin'];
      this.tokenService.generateAccessToken({ did: userDid, verifiedRoles, origin }),
```

src/modules/auth/token.service.ts
jrhender: I don't really understand how the `HOST` header is relevant here. Isn't the `HOST` the destination of the request? So I don't understand how it says anything about the client.

src/modules/auth/token.service.ts
```diff
@@ -64,4 +69,28 @@ export class TokenService {
...
+
+    if (req.headers['authorization']) {
+      token = req.headers['authorization'].replace('Bearer ', '');
+    } else {
+      token = req.cookies.token
+    }
+
+    if (token) {
+      const decodedToken = this.jwtService.decode(token) as TokenPayload;
+      if (decodedToken.supportedOrigins.includes(requestOrigin) ||
```
jrhender, line 85: A continuation of the implementation on my first comment, I think this is better logic:
```suggestion
      if (decodedToken.origin === requestOrigin || requestOrigin === undefined)
```
The first case is for browsers which are setting the origin header and the second case is for servers which are not setting the origin header. Does this make sense?

jrhender (changes requested):

src/modules/auth/token.service.ts
```diff
@@ -70,9 +69,8 @@ export class TokenService {
     return this.refreshTokenRepository.deleteRefreshTokenById(id);
   }
 
-  async handleOriginCheck(req: Request, res: Response, next: NextFunction) {
+  async handleOriginCheck(req: Request, res: Response, next: NextFunction, service: JwtService) {
```
jrhender, line 72: Please add some documentation to explain why we are checking the origin (i.e. for csrf protection) and why we have chosen this approach

.6e207494f2ab9f0e3cabc073301297a9300fc69f-audit.json
```diff
@@ -9,6 +9,11 @@
             "date": 1625476684155,
             "name": "iam-cache-server-2021-07-05.log",
             "hash": "6844220f4c1420a7133c65934580eee7"
+        },
+        {
+            "date": 1625555833274,
+            "name": "iam-cache-server-2021-07-06.log",
+            "hash": "1413b3de62f75a2c770b5095588a1f4b"
         }
     ]
```
jrhender: This file doesn't need to be committed. Could you add to the `.gitignore` so that isn't added please?

src/modules/auth/token.service.ts
```diff
@@ -81,10 +79,8 @@ export class TokenService {
     }
 
     if (token) {
-      const decodedToken = this.jwtService.decode(token) as TokenPayload;
-      if (decodedToken.supportedOrigins.includes(requestOrigin) ||
-        decodedToken.supportedHosts.includes(req.hostname) // swagger and server based requests check only
-      ) {
+      const decodedToken = service.decode(token) as TokenPayload;
+      if (decodedToken.origin === req.headers['origin'] || req.headers['origin'] === undefined) {
```
jrhender, line 83: Would be good to explain why the logic is as it is
```suggestion
      const isBrowserRequestFromAuthenticatedOrigin = decodedToken.origin === req.headers['origin'] 
      const isServerRequestOrGETFromSameDomain =  req.headers['origin'] === undefined
      if (isBrowserRequestFromAuthenticatedOrigin || isServerRequestOrGETFromSameDomain) {
```

jrhender (changes requested):
@hrachocode Please add some tests covering this functionality. If you need help @ahmedolaibrahim is knowledgable with nestjs tests

src/modules/auth/token.service.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+        const module: TestingModule = await Test.createTestingModule({
+            imports: [TestModule],
+            controllers: [TestController],
+        }).compile();
+
+        app = module.createNestApplication();
+        await app.init();
+    });
+
+    it('should have same GET request ORIGIN header as stored in JWT token ', async () => {
```
jrhender, line 74: Is the `origin` header set for this test? I think it is probably not being set as it is a node request? If so, I think this test name is misleading

src/modules/auth/token.service.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+            .apply((req, res, next) => this.tokenService.handleOriginCheck(req, res, next))
+            .exclude(
+                { path: '/login', method: RequestMethod.ALL },
+                { path: '/refresh_token', method: RequestMethod.ALL }
+            )
+            .forRoutes({ path: '/*', method: RequestMethod.ALL });
+    }
+};
+
+describe('TokenService', () => {
```
jrhender, line 61: Would it be possible to add a test that demonstrates that the token check fails if the origin header doesn't match the token? Or is that not possible because it's not possible to set the origin header in the test?

jrhender (commented):
Looks like tests are coming along 👍  
Some documentation on `handleOriginCheck` would still be nice please 😄  Otherwise, that rationale for our CSRF approach won't be clear to future developers, IMO

src/modules/auth/token.service.ts
```diff
@@ -64,4 +68,26 @@ export class TokenService {
...
+      const decodedToken = service.decode(token) as TokenPayload;
+      const isBrowserRequestFromAuthenticatedOrigin = decodedToken.origin === req.headers['origin']
+      const isServerRequestOrGETFromSameDomain = req.headers['origin'] === undefined
+      if (isBrowserRequestFromAuthenticatedOrigin || isServerRequestOrGETFromSameDomain) {
+        next()
+      } else {
+        throw new InternalServerErrorException('Something went wrong')
+      }
+    } else {
+      throw new UnauthorizedException('Unauthorized')
```
ahmedolaibrahim, line 90:
```suggestion
      throw new UnauthorizedException('Unauthorized');
```

src/modules/auth/token.service.spec.ts
ahmedolaibrahim: +1

ahmedolaibrahim (commented):
Looking good @hrachocode  just some few improvements  needed.

src/modules/auth/token.service.ts
ahmedolaibrahim: looks like there are instances where statements are terminated with semi-colon and some are not. let make them consistent.

src/modules/auth/token.service.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+            const isServerRequestOrGETFromSameDomain = req.headers['origin'] === undefined
+            if (isBrowserRequestFromAuthenticatedOrigin || isServerRequestOrGETFromSameDomain) {
+                next()
+            } else {
+                throw new InternalServerErrorException('Something went wrong')
+            }
+        } else {
+            throw new UnauthorizedException('Unauthorized')
+        }
+    }
```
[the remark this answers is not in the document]
hrachocode: the reason behind this is when we are using an imported function from TokenService we should also import all other dependencies and their configs including AuthModule for example. 

so this way we are creating a mock of that function rather then working with dependencies.

src/modules/auth/token.service.spec.ts
hrachocode: sure its possible just by setting  `.set({ 'origin': 'anydomain.com' })` [here](https://github.com/energywebfoundation/iam-cache-server/blob/3e8a70a081fa209b3106761955cd40cc9e47a70e/src/modules/auth/token.service.spec.ts#L79), so if origin in headers will no match with origin in token it will throw 500 internal server error.

src/modules/auth/token.service.spec.ts
hrachocode: for the current version, the origin header is undefined which also matches  our check, will add an origin with headers so naming should be consistent then.

src/modules/auth/token.service.ts
hrachocode: sure, thank you for suggestion

src/modules/auth/auth.module.ts
```diff
@@ -36,6 +36,18 @@ import { ConfigService } from '@nestjs/config';
...
-  exports: [JwtAuthGuard, JwtStrategy, GqlAuthGuard],
+  exports: [JwtAuthGuard, JwtStrategy, GqlAuthGuard, TokenService],
 })
-export class AuthModule {}
+export class AuthModule implements NestModule {
+  constructor(private readonly tokenService: TokenService, private readonly jwtService: JwtService) { }
+
+  configure(consumer: MiddlewareConsumer) {
+    consumer
+      .apply((req, res, next) => this.tokenService.handleOriginCheck(req, res, next, this.jwtService))
```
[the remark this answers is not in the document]
hrachocode: agree, removed, thank you.

src/modules/auth/token.service.ts
hrachocode: thank you for the note, check in the next commit in docs/security folder.

src/modules/auth/token.service.spec.ts
jrhender: As a stretch goal, it would be nice to have tests of other HTTP methods (e.g. `POST`)

src/modules/auth/token.service.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+            if (isBrowserRequestFromAuthenticatedOrigin || isServerRequestOrGETFromSameDomain) {
+                next()
+            } else {
+                throw new InternalServerErrorException('Something went wrong');
+            }
+        } else {
+            throw new UnauthorizedException('Unauthorized');
+        }
+    }
+}
```
jrhender, line 34: @hrachocode you wrote:
> the reason behind this is when we are using an imported function from TokenService we should also import all other dependencies and their configs including AuthModule for example. so this way we are creating a mock of that function rather then working with dependencies.

Sorry, I'm not understanding this at all 😅 . In this case, the `TokenService` is the class under test, so it definitely shouldn't be mocked. If the test is using a mock `TokenService` then how does the test tell us anything about `TokenService` that is actually used during runtime by the app?

src/modules/auth/token.service.ts
jrhender: Nice, the docs file looks good. IMO, it would be better to actually put the documentation next to the class itself as JSDoc to make it more likely that it will be read an maintained. This also allows the documentation to be read as intellisense.
I also added a line at the end about why other patterns couldn't be used
```suggestion
/**
 * Our approach to prevent or at least maximum decrease chances for any CSRF attacks:
 *
 * Provide a Protection using Origin in headers and combining double check with JWT: 
 * * Origin is a request header supported in most modern browsers, indicating from where the request originated, **which cannot be modified** <https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Origin>
 * * We will store the request Origin header in JWT token after login
 * * And with every request our middleware will check if Origin in request headers is matching with origin stored in JTW token(**hacker cannot modify JWT as we are signing it with our secret in backend**)
 * * For node requests we will check if origin in request headers is undefined. 
 * * IF ORIGIN IS SAME OR UNDEFINED: everything is fine
 * * IF NO: throw an exception as the origin could be from a hacker's website.
 * * More info(check approach 2, **origin headers**): <https://security.stackexchange.com/questions/203890/how-to-implement-csrf-protection-with-a-cross-origin-request-cors/203910#203910>
 *
 * A pattern such as the double cookie submit pattern cannot be used because the cache-server does not share an origin with it's clients.
 */
  async handleOriginCheck(req: Request, res: Response, next: NextFunction, service: JwtService) {
```

docs/security/csrfProtection.md
```diff
@@ -0,0 +1,15 @@
+Our approach to prevent or at least maximum decrease chances for any CSRF attacks
```
jrhender, line 1: Documentation looks good but this file can be removed if we go with JSDoc next to the method itself

jrhender (changes requested):
Getting there but still a couple points to address, in particular the use of a "mocked" `TokenService` in the tests

src/modules/auth/auth.module.ts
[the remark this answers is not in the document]
hrachocode: oh, sorry for this, I thought I need to remove TokenService from exports, as we added it for testing. now also committing the needed change to remove TokenService from handleORiginCheck arguments and using jsonwebtoken instead.

src/modules/auth/token.service.spec.ts
hrachocode: updated by latest [commit](https://github.com/energywebfoundation/iam-cache-server/pull/145/commits/34318563453d02a5a399f48163b79890cb244c15)

src/modules/auth/token.service.spec.ts
hrachocode: updated by latest [commit](https://github.com/energywebfoundation/iam-cache-server/pull/145/commits/34318563453d02a5a399f48163b79890cb244c15)

src/modules/auth/auth.module.ts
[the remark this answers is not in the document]
hrachocode: updated by latest [commit](https://github.com/energywebfoundation/iam-cache-server/pull/145/commits/34318563453d02a5a399f48163b79890cb244c15)

docs/security/csrfProtection.md
hrachocode: done

src/modules/auth/token.service.spec.ts
hrachocode: Ok in other words, when removing the Mock function and using `TokenService` class as a provider to our test module
it throws errors like 

`Nest can't resolve dependencies of the TokenService (?, JwtService, RefreshTokenRepository). Please make sure that the argument ConfigService at index [0] is available in the TestModule context.`

and when I am trying to import `ConfigService` in `TestModule` providers, then it requires `RefreshTOkenRepository`, after it requires `JwtService`

and after it triggers an error that `JWT_MODULE_OPTIONS` are not provided.

and this list is deeper as when I am trying to use real `TokenService class`  :/ 

that's why I decided to mock that function rather than importing all project configs in one test file.

jrhender: As I understand, mocking the dependencies of the class under test is the whole point of mock. We test the real class but mock its dependencies. So instead of importing the real `ConfigService` (which then requires other imports as you say), we instead mock `ConfigService` so that it doesn't need any imports

hrachocode: yeah correct, but we are mocking `TokenService` not `ConfigService`

jrhender (approved):
Looks good 🎉

PR #146, opened 2021-07-06
Feature/swtch 673 api versioning

hrachocode:


src/modules/logger/logger.service.ts
```diff
@@ -52,6 +52,10 @@ export class Logger extends NestLogger implements LoggerService {
     });
   }
 
+  setContext(ctx: string) {
+    this.context = ctx;
+  }
```
jrhender, line 57: Why this change?

watch
```diff
@@ -0,0 +1,4 @@
+
+> iam-cache-server@0.0.1 test /home/hrachocode/energy_web/iam-cache-server
+> jest
+
```
jrhender, line 4: Why is this file committed?

src/modules/assets/assets.controller.ts
```diff
@@ -20,10 +22,11 @@ import { Order } from './assets.types';
 @Auth()
 @UseInterceptors(SentryErrorInterceptor)
 @ApiTags('Assets')
-@Controller('assets')
+@Controller({ path: 'assets', version: VERSION_NEUTRAL })
 export class AssetsController {
   constructor(private readonly assetsService: AssetsService) { }
 
+  @Version('1')
```
jrhender, line 29: Why only this route is version 1?

.6e207494f2ab9f0e3cabc073301297a9300fc69f-audit.json
```diff
@@ -0,0 +1,19 @@
...
+            "name": "iam-cache-server-2021-06-28.log",
+            "hash": "10d97ae90fcfec091eb57b9dd73f90dd"
+        },
+        {
+            "date": 1625560590093,
+            "name": "iam-cache-server-2021-07-06.log",
+            "hash": "5cfb23c81530b1088ea90646c1234043"
+        }
+    ]
+}
```
jrhender, line 19: don't commit. please add to gitignore

jrhender (commented):
@hrachocode Looks good. Are there any other changes that you think we still need to make? (I just noticed that the PR is in draft)

A question for you:
- how does the Swagger page look if there are two different versions of a controller. I found this SO question but it is for .NET: https://stackoverflow.com/questions/60847380/create-2-different-swagger-ui-for-that-which-can-run-with-two-different-version

src/modules/logger/logger.service.ts
hrachocode: after migrating to nest js v8.0.0-alpha(not sure if thats the reason), I started getting this kinda error in all services where Logger is used

image

I added this method to sets the context of Logger, accepting context name.

it was missing and can't figure out how it was working before

watch
hrachocode: added to gitignore

src/modules/assets/assets.controller.ts
hrachocode: sorry, this was for testing purposes, removed.

hrachocode:
> @hrachocode Looks good. Are there any other changes that you think we still need to make? (I just noticed that the PR is in draft)
> 
> A question for you:
> 
> * how does the Swagger page look if there are two different versions of a controller. I found this SO question but it is for .NET: https://stackoverflow.com/questions/60847380/create-2-different-swagger-ui-for-that-which-can-run-with-two-different-version

Interesting question, as far as I understand we should serve different swagger versions based on some swagger configs, for v1 v2 v3, etc. I think Nest js is preparing an update for nestjs/swagger v8 so it will support some options for API versioning as well

https://github.com/nestjs/nest/issues/5065#issuecomment-874796808

jrhender:
@hrachocode could you confirm that you can login to Switchboard if running against this cache-server branch? I'm getting this error: 
```
[Nest] 2224  - 07/07/2021, 11:50:30 AM   ERROR [ExceptionsHandler] rxjs_1.lastValueFrom is not a function
TypeError: rxjs_1.lastValueFrom is not a function
    at RouterResponseController.transformToResult (/Users/john/ew/iam-cache-server/node_modules/@nestjs/core/router/router-response-controller.js:32:27)
    at /Users/john/ew/iam-cache-server/node_modules/@nestjs/core/router/router-execution-context.js:173:52
    at /Users/john/ew/iam-cache-server/node_modules/@nestjs/core/router/router-execution-context.js:47:19
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
    at /Users/john/ew/iam-cache-server/node_modules/@nestjs/core/router/router-proxy.js:9:17
```
image

hrachocode:
Hmmm ,very strange error, didnt get This kind of, will test again tomorrow

JGiter (changes requested):
There are unmet peer dependencies updated packages. Is it possible to resolve them?

jrhender:
@hrachocode I think this PR can be closed now, right?

hrachocode:
Agree

PR #150, opened 2021-08-02
Bug/ics-18 ens app sync failure

ahmedolaibrahim:
https://energyweb.atlassian.net/browse/ICS-18


JGiter (changes requested):
Could you also test synchronization of namespace with empty owner or malformed metadata?

src/modules/ens/ens.service.unit.spec.ts
```diff
@@ -122,6 +122,21 @@ describe('EnsService', () => {
     await app.close();
   });
 
+  describe('Sync ENS', () => {
+    it('syncENS() throw an error', async () => {
```
JGiter, line 126: It is a bit vague name, because synchronization can throw by many reasons. Would it be more descriptive `deleted namespace synchronization should throw`?

src/modules/ens/ens.service.unit.spec.ts
```diff
@@ -0,0 +1,256 @@
...
+    }, 30000);
+
+    it('syncNamespace() Validate and accept ROLE Sync Data with roleType app', async () => {
+      const mockRoleServiceSpy = jest.spyOn(
+        MockRoleService,
+        'handleRoleSyncWithEns',
+      );
+      await service.syncNamespace({
+        data: ROLE_MOCK_DATA,
+        namespace: 'test.roles.onion.iam.ewc',
```
JGiter, line 186: Since role type is `app` this role must be in application domain

JGiter (changes requested):

src/modules/ens/ens.service.ts
```diff
@@ -141,7 +141,7 @@ export class EnsService {
     });
   }
 
-  private async getAllNamespaces() {
+  async getAllNamespaces() {
```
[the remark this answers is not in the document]
ahmedolaibrahim: yes that's correct.

src/modules/ens/ens.service.ts
```diff
@@ -159,17 +159,17 @@ export class EnsService {
     owner?: string;
   }) {
     try {
-
+      if (!owner  || owner !== emptyAddress) {// prevents attempt to resync after namepsace deregisteration
```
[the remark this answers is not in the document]
ahmedolaibrahim: The reason for allowing synchronisation when there is no owner is that the `owner` param is optional(like you mentioned the presence of the owner param depends on the event ) and so makes sense for it to try to resolve that below [here](https://github.com/energywebfoundation/iam-cache-server/blob/1a344c61e64b606d255cf935cd0021112b137b48/src/modules/ens/ens.service.ts#L257).

What namespace should be removed? I think there is a function to handle that already  removal of namespace from the db and this method isn't responsible for that so I'm not sure what duplication you are referring to.

This only ensures that namespaces that have been removed do not attempt to resync again because it throws an error.

src/modules/ens/ens.service.unit.spec.ts
ahmedolaibrahim: can you be more specific about this? are you saying it should be `test.roles.onion.app.iam.ewc`?

jrhender: I think he is saying that it should be `test.roles.myapp.apps.onion.iam.ewc`

jrhender (approved):
Looks good to me, nice tests 👍

src/modules/ens/ens.service.ts
```diff
@@ -159,7 +159,7 @@ export class EnsService {
     owner?: string;
   }) {
     try {
-      if (!owner  || owner !== emptyAddress) {// prevents attempt to resync after namepsace deregisteration
+      if (owner !== emptyAddress) {// prevents attempt to resync after namepsace deregisteration
```
JGiter: Won't it prevent deletion of the domain? To delete domain we set its owner to empty address, but this check just stops processing instead of removing domain from db. Also since e2e testing already configured we could make use of it here

JGiter (changes requested):

src/modules/ens/ens.service.ts
```diff
@@ -108,6 +108,32 @@ export class EnsService {
  
   }
 
+  private async namespaceDeleteHandler(name: string) {
```
JGiter, line 111: I suggest more straightforward `deleteNamespace`

JGiter (approved):

PR #152, opened 2021-08-10
Feat/pda 4 use versioned api

jrhender:


src/modules/logger/logger.service.ts
```diff
@@ -52,6 +52,10 @@ export class Logger extends NestLogger implements LoggerService {
     });
   }
 
+  setContext(ctx: string) {
+    this.context = ctx;
+  }
+
```
hrachocode, line 58: so when using a custom logger we should set the context manually? or asking other way, I want to understand the purpose of this change?

jrhender: Haha, this is the same question I asked on your PR, I think: https://github.com/energywebfoundation/iam-cache-server/pull/146#discussion_r665036124
And the answer is the same. I guess `NestLogger` must have dropped their implementation of `setContext` in v8

jrhender:
@knzeng-e we should squash the commit in this PR before merging. We've got some small commits that we don't need in the history.

src/modules/logger/logger.service.ts
hrachocode: ah:D:D forgot about that:D yeah then its fine from my side<3

jrhender:
> > @knzeng-e we should squash the commit in this PR before merging. We've got some small commits that we don't need in the history.
> 
> Yes I agree @jrhender !
> is it ok if we keep those ?
> 
> * from [3cbb175](https://github.com/energywebfoundation/iam-cache-server/commit/3cbb175520634f6ec91cab19733de532255edfc7) (upgrade nest.js to v8) to [35f26d6](https://github.com/energywebfoundation/iam-cache-server/commit/35f26d66af7dfe7249d820dbebce67c3a3a56b98) (fix Logguer import)
> * [c2f9fcf](https://github.com/energywebfoundation/iam-cache-server/commit/c2f9fcf532ef6df24ccb2f748e3d4f32dc50de5b) (setting version neutral api routes)
> * [c79db1c](https://github.com/energywebfoundation/iam-cache-server/commit/c79db1cfb2ce0c12191b2c6c791e50cd7e3939c3) (The second install and use nestjs/axios for HttpModule)

@knzeng-e perfect, thanks!

PR #153, opened 2021-08-11
Fix queryFailedError relating to missing ClaimIssuer column

ahmedolaibrahim:
- [x] Removes where condition using `claimIssuer` Column in `getByRequestermethod `
- [x] Removes where condition using `claimIssuer` Column in `getUserByDid `

### DIFF
--- a/src/modules/claim/claim.service.ts
+++ b/src/modules/claim/claim.service.ts
@@ -24,7 +24,7 @@ interface QueryFilters {
   namespace?: string;
 }
 
-const UUID_NAMESPACE = "5193850c-2367-4ec4-8c22-95dfbd4a2880";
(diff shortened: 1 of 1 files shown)


src/modules/claim/claim.service.ts
```diff
@@ -223,11 +228,11 @@ export class ClaimService {
     }
 
     if (currentUser) {
-      qb.andWhere(new Brackets(query => {
-        query.where(':currentUser = ANY ("claimIssuer")', {
-          currentUser,
-        }).orWhere(':currentUser = requester', { currentUser })
```
jrhender, line 229: Seems like this is the only substantive change in the PR?

jrhender (approved):
Looks good to me 👍

PR #154, opened 2021-08-18
bug/ICS-43 can't get subject claims

JGiter:
https://energyweb.atlassian.net/jira/software/projects/ICS/boards/86?selectedIssue=ICS-43
Additionally 

- closed Nats connection on Nats module destroy, because it is prevents jest from exit
- configure redis for CI


ahmedolaibrahim (commented):
looks good.. I think we should add some tests for this though

ahmedolaibrahim (approved):

PR #155, opened 2021-08-18
feat/ics-41 persist staking pool terms and conditions

ahmedolaibrahim:
- [x] Persist staking pool terms and conditions
- [x] Create event listeners for staking pool launch event


src/modules/ens/ens.service.ts
```diff
@@ -21,8 +21,12 @@ import {
   DomainHierarchy,
   ResolverContractType,
   DomainNotifier__factory,
+  StakingPoolFactory__factory,
+  VOLTA_STAKING_POOL_FACTORY_ADDRESS
 } from '@energyweb/iam-contracts';
 import type { DomainNotifier } from "@energyweb/iam-contracts/dist/ethers-v4/DomainNotifier";
+import { StakingService } from '../staking/staking.service';
+import { StakingPoolFactory } from '@energyweb/iam-contracts/dist/ethers-v4/StakingPoolFactory';
```
jrhender, line 29: Why is Staking being mixed in with the ENS code? In my opinion, it is important to try to keep this functionality separate. Ideally, it would be possible to split out a "staking" cache service from the "ens" cache service, IMO.

JGiter (changes requested):
Are you sure that all pools should have same terms? I thought that service provider may determine at least portion of patron share

JGiter:
Although tests marked as passed there are unhandled promise rejection in staking tests as well as in organization tests

ahmedolaibrahim:
Screenshot 2021-08-25 at 12 57 35
@JGiter here's a screenshot attached showing the tests being run without staking or organisation tests and the unhandled promise rejection warnings are still present so it isn't those tests causing it. 
I have however created a story to handle this [here](https://energyweb.atlassian.net/browse/ICS-45)

JGiter (changes requested):
Could you please elaborate on some details? On `StakingPool` set one-to-one relation with `StakingTerms` entity, but `saveStakingPool` assigns latest version of terms to new pool, which means many-to-one relation. Also since terms can't exist by their own, I believe terms should be updated or read by pool which they are belongs to.

JGiter (approved):

PR #156, opened 2021-08-31
Release to Volta

jrhender:
Volta Oct 2021 Release


src/modules/claim/claim.controller.ts
```diff
@@ -241,16 +264,47 @@ export class ClaimController {
   })
   public async getByRequesterDid(
     @Param('did') requester: string,
-    @Query('accepted', IsAcceptedPipe)
-    accepted?: boolean,
-    @Query('namespace') parentNamespace?: string,
+    @Query('isAccepted', BooleanPipe) isAccepted?: boolean,
+    @Query('namespace') namespace?: string,
     @User() user?: string,
```
jrhender, line 269: Example breaking change to API

PR #157, opened 2021-09-05
Bug/ICS-22 cleanup duplicate namespaces

ahmedolaibrahim:
- [x]  CleanUp Duplicate ENS namespaces
- [x] Prevent Duplicates 

To Test The Duplicates Removal

- Create a dump of staging DB
- Create a fresh DB locally and import dump from staging
- Run migration Command

### DIFF
--- a/src/migrations/1630744507207-RemoveDuplicates.ts
+++ b/src/migrations/1630744507207-RemoveDuplicates.ts
@@ -1,0 +1,44 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+const deleteDuplicateOrgs = async (queryRunner: QueryRunner) => {
+  await queryRunner.query(
+    `DELETE FROM organization org
+            WHERE  EXISTS (
+            SELECT FROM organization
+            WHERE  namespace = org.namespace
+            AND    id < org.id
+        );`,
+  );
+};
+
+const deleteDuplicateApps = async (queryRunner: QueryRunner) => {
+  await queryRunner.query(
+    `DELETE FROM application app
+            WHERE  EXISTS (
+            SELECT FROM application
+            WHERE  namespace = app.namespace
+            AND    id < app.id
+        );`,
+  );
+};
+
+const deleteDuplicateRoles = async (queryRunner: QueryRunner) => {
+  await queryRunner.query(
+    `DELETE FROM role r
+            WHERE  EXISTS (
+            SELECT FROM role
+            WHERE  namespace = r.namespace
+            AND    id < r.id
+        );`,
+  );
+};
+
+export class RemoveDuplicates1630744507207 implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await deleteDuplicateOrgs(queryRunner);
+    await deleteDuplicateApps(queryRunner);
+    await deleteDuplicateRoles(queryRunner);
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {}
+}
--- a/src/migrations/1630744507207-RemoveDuplicates.ts
+++ b/src/migrations/1630744507207-RemoveDuplicates.ts
@@ -40,5 +40,6 @@ export class RemoveDuplicates1630744507207 implements MigrationInterface {
     await deleteDuplicateRoles(queryRunner);
   }
 
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
   public async down(queryRunner: QueryRunner): Promise<void> {}
 }
--- a/src/modules/application/application.service.spec.ts
+++ b/src/modules/application/application.service.spec.ts
@@ -1,19 +1,132 @@
+import { ConfigModule } from '@nestjs/config';
 import { Test, TestingModule } from '@nestjs/testing';
+import {
+  getRepositoryToken,
+  TypeOrmModule,
+  TypeOrmModuleOptions,
+} from '@nestjs/typeorm';
+import { Application } from '../application/application.entity';
+import { LoggerModule } from '../logger/logger.module';
+import { SentryModule } from '../sentry/sentry.module';
+import * as TestDbCOnfig from '../../../test/config';
+import { Connection, EntityManager, QueryRunner, Repository } from 'typeorm';
+import { Chance } from 'chance';
+import { Logger } from '../logger/logger.service';
 import { ApplicationService } from './application.service';
+import { OrganizationService } from '../organization/organization.service';
+import { organizationFixture } from '../organization/organization.fixture';
+import { Organization } from '../organization/organization.entity';
+import { applicationFixture } from './application.fixture';
 
-// TODO: fix test so pending can be removed
-xdescribe('ApplicationService', () => {
+const chance = new Chance();
+
+const MockLogger = {
+  log: jest.fn(),
+  error: jest.fn(),
+  setContext: jest.fn(),
+  info: jest.fn(),
+  debug: jest.fn(),
+};
+
+const MockOrgService = {
+  getByNamespace: jest.fn(),
+};
+
+describe('ApplicationService', () => {
   let service: ApplicationService;
+  let module: TestingModule;
+  let repo: Repository<Application>;
+  let orgRepo: Repository<Organization>;
+  let organizations: Organization[];
+  let applications: Application[];
+  let queryRunner: QueryRunner;
 
   beforeEach(async () => {
-    const module: TestingModule = await Test.createTestingModule({
-      providers: [ApplicationService],
+    module = await Test.createTestingModule({
+      imports: [
+        TypeOrmModule.forRoot(TestDbCOnfig.default as TypeOrmModuleOptions),
+        TypeOrmModule.forFeature([Organization, Application]),
+        LoggerModule,
+        ConfigModule,
+        SentryModule,
+      ],
+      providers: [
+        ApplicationService,
+        { provide: OrganizationService, useValue: MockOrgService },
+        {
+          provide: Logger,
+          useValue: MockLogger,
+        },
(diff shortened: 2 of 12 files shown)


src/migrations/1630744507207-RemoveDuplicates.ts
```diff
@@ -0,0 +1,45 @@
...
+const deleteDuplicateApps = async (queryRunner: QueryRunner) => {
+  await queryRunner.query(
+    `DELETE FROM application app
+            WHERE  EXISTS (
+            SELECT FROM application
+            WHERE  namespace = app.namespace
+            AND    id < app.id
+        );`,
+  );
+};
```
ahmedolaibrahim, line 23: Duplicates Apps found in the database all turned out to have the same `parentOrgId` of the first OrgId created before other duplicate orgs hence it makes sense for us to remove duplicates except the first one created.

src/migrations/1630744507207-RemoveDuplicates.ts
```diff
@@ -0,0 +1,45 @@
...
+const deleteDuplicateRoles = async (queryRunner: QueryRunner) => {
+  await queryRunner.query(
+    `DELETE FROM role r
+            WHERE  EXISTS (
+            SELECT FROM role
+            WHERE  namespace = r.namespace
+            AND    id < r.id
+        );`,
+  );
+};
```
ahmedolaibrahim, line 34: same reason as above

src/migrations/1630744507207-RemoveDuplicates.ts
```diff
@@ -0,0 +1,45 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+const deleteDuplicateOrgs = async (queryRunner: QueryRunner) => {
+  await queryRunner.query(
+    `DELETE FROM organization org
```
ahmedolaibrahim, line 5: removes all duplicate except the first created org.

src/modules/organization/organization.service.ts
```diff
@@ -97,8 +97,18 @@ export class OrganizationService {
...
-    return Boolean(await this.getByNamespace(namespace));
+  public async exists(
+    namespace: string,
+    parentOrg?: Organization,
+  ): Promise<boolean> {
+    const whereObjectClause: any = { namespace };
+    if (parentOrg) {
+      whereObjectClause.parentOrg = parentOrg;
+    }
+    const appExists = await this.orgRepository.findOne({
```
JGiter, line 108:
```suggestion
    const orgExists = await this.orgRepository.findOne({
```

src/modules/role/role.service.ts
```diff
@@ -46,8 +48,26 @@ export class RoleService {
    * return true if role with given namespace exists
    * @param namespace
    */
-  public async exists(namespace: string) {
-    return Boolean(await this.getByNamespace(namespace));
+  public async exists(
```
JGiter, line 51: This signature does not allow to check only parent organization

src/modules/role/role.fixture.ts
```diff
@@ -0,0 +1,47 @@
...
+      metadata: {},
+    };
+
+    const role = Role.create({
+      name,
+      namespace: `${name}.roles.testapp.apps.testApp.iam.ewc`,
+      owner: '0x7dD4cF86e6f143300C4550220c4eD66690a655fc',
+      definition,
+      parentOrg: organization,
+      parentApp: application,
```
JGiter, line 41: Role has only one parent which is app or org

JGiter (changes requested):

src/modules/application/application.service.spec.ts
```diff
@@ -1,19 +1,127 @@
...
-  it('should be defined', () => {
-    expect(service).toBeDefined();
+  afterEach(async () => {
+    await queryRunner.rollbackTransaction();
+    module.close();
+  });
+
+  it('create() it should not save an application with existing namespace and parentOrg ', async () => {
+    const testApp = chance.pickone(applications);
+    const organization = chance.pickone(organizations);
```
jrhender, line 85: Why the use of `chance` here? I tend to think tests should not have randomness so that they are repeatable, but maybe I don't understand the rationale here

src/modules/application/application.service.spec.ts
```diff
@@ -1,19 +1,127 @@
...
+      owner,
+      definition,
+      parentOrg: organization.namespace,
+    });
+
+    expect(MockLogger.debug).toHaveBeenCalledWith(
+      expect.stringContaining(
+        `Not able to create application: ${namespace} with same parent organization ${organization.namespace} already exists`,
+      ),
+    );
```
jrhender, line 102: I think testing the logger is a find approach, but maybe it would be even better to query the db and ensure that there aren't duplicates using the `service`?

src/modules/application/application.service.ts
```diff
@@ -60,8 +61,18 @@ export class ApplicationService {
...
+    parentOrg?: Organization,
+  ): Promise<boolean> {
+    const whereObjectClause: any = { namespace };
+    if (parentOrg) {
+      whereObjectClause.parentOrg = parentOrg;
+    }
+    const appExists = await this.applicationRepository.findOne({
+      where: whereObjectClause,
+    });
+    return Boolean(appExists);
```
jrhender, line 75: Why this change? Is it fixing a bug or just adding the ability to also have a check for the `parentOrg`. I would think that the `parentOrg` is implied in the namespace already. Like if a namespace like `finance.apps.ahmed.iam.ewc` is passed a `namespace`, aren't we already checking that the parent is `ahmed.iam.ewc` implicitly 🤔 ?

src/modules/application/application.service.ts
```diff
@@ -60,8 +61,18 @@ export class ApplicationService {
    * return true if App with given namespace exists
    * @param namespace
    */
```
jrhender, line 63: Missing documentation for new `parentOrg` param

src/modules/role/role.service.ts
```diff
@@ -46,8 +48,26 @@ export class RoleService {
    * return true if role with given namespace exists
    * @param namespace
    */
```
jrhender, line 50: Missing documentation

src/modules/organization/organization.service.ts
```diff
@@ -97,8 +97,18 @@ export class OrganizationService {
    * return true if Org with given namespace exists
    * @param namespace
    */
```
jrhender, line 99: Missing documentation

src/modules/role/role.fixture.ts
jrhender: I agree with Dmitry. This should be enforced in `Role.create`, I believe

jrhender (changes requested):
Some questions and missing documentation

src/modules/application/application.service.spec.ts
ahmedolaibrahim: Yeah I agree with you that tests should not have randomness but in this case, using chance to pick one of an already created app or org doesn't affect the outcome of the tests because what is selected is what is going to be saved at the end and that is what we are asserting for.

src/modules/application/application.service.spec.ts
ahmedolaibrahim: Not sure that would be useful in this case though, remember we are trying to test if we are able to save or not  and the Logger is only called when we are not able to save it. querying the DB would be an additional call we don't actually need.

src/modules/application/application.service.ts
ahmedolaibrahim: While cleaning up duplicates, I discovered that, There are cases where we have duplicates with the same `namespace` and `parentOrg` so this is more like a check for if that namespace with that parentOrg already exists.

src/modules/role/role.service.ts
ahmedolaibrahim: This exists method is for the `roles`  i don't think there's a reason for us to be checking for only a parentOrg here, we can either check only for a role namespace or role namespace belonging to a parentOrg. we can check for only parent org namespace in the exists method for organisation.

JGiter: I mean that optional arguments can be omitted, but `parentApp` can't be omitted, when you want to filter only by `parentOrg`

src/modules/application/application.service.spec.ts
jrhender: Hmm, I was suggesting querying the DB in the test to ensure that we don't have a duplicate because that ultimately what we care about. Is that was you're talking about too?
My rationale is that, I know it's not currently doing this, but it would be possible to log that message but still somehow create a duplicate in the DB, in which case the test would be misleading IMO.

src/modules/application/application.service.ts
jrhender: Hmm, okay, I'm still not fully understanding when someone would use the `parentOrg` param. What I mean is that, if I'm querying for a role like `engineer.roles.ahmed.iam.ewc`, the parent org can only be `ahmed.iam.ewc`. So if someone were to make a query like `exists(engineer.roles.ahmed.iam.ewc, john.iam.ewc)`, that is obviously going to be false. Am I misunderstanding something here?

ahmedolaibrahim: Sorry for the misunderstanding. you're right. Dmitry clarified this to me this morning.

src/modules/application/application.service.spec.ts
ahmedolaibrahim: okay make sense.

ahmedolaibrahim:
@jrhender @JGiter Closing this in favour of [this](https://github.com/energywebfoundation/iam-cache-server/pull/158). the other PR contains changes requested here and  namespace deletion sync.

PR #158, opened 2021-09-08
Bug/ICS-8 Namespace Sync

ahmedolaibrahim:
- [x] Cleanup Duplicates in database
- [x] Prevents Duplicates in database
- [x] Sync Namespace deletion

This PR includes changes requested and feature changes in this other [PR](https://github.com/energywebfoundation/iam-cache-server/pull/157) as well as functionality for syncing namespace deletion. The were initially meant to be separate PRs but for some reason I branched off the other PR instead of master.


src/modules/application/application.service.ts
```diff
@@ -110,18 +129,33 @@ export class ApplicationService {
     return this.applicationRepository.delete(app.id);
   }
 
+  /**
+   * removes App with matching namehash
+   * @param namehash
+   */
+  public async removeByNameHash(namehash: string) {
+    const app = await this.getByNamehash(namehash);
```
JGiter, line 137: Wouldn't it be more performant to send only one delete query with WHERE clause?

src/migrations/1631103259350-AddNamehashColumn.ts
```diff
@@ -0,0 +1,25 @@
...
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `ALTER TABLE "role" ADD "namehash" character varying`,
+    );
+    await queryRunner.query(
+      `ALTER TABLE "organization" ADD "namehash" character varying`,
+    );
+    await queryRunner.query(
+      `ALTER TABLE "application" ADD "namehash" character varying`,
+    );
```
jrhender, line 15: Not a big deal, but the `namehash` actually isn't character varying. It is a fixed 256bit (or 64 hex characters, 66 with the `0x` prefix), so the column size could be fixed as well. Not sure if makes sense to do this though 🤷  https://dba.stackexchange.com/a/21496 . So probs best to leave it character varying 🤔 ? what do you think @ahmedolaibrahim ?

src/modules/application/application.entity.ts
```diff
@@ -49,6 +49,10 @@ export class Application implements BaseEnsEntity {
   @Column()
   namespace: string;
 
+  @Field({ nullable: true })
+  @Column({ nullable: true })
```
jrhender, line 53: I think we need this to be `nullable` to start as some records won't have it populated. Or can we not have the `nullable` from the get go? 
```suggestion
```
Either way, we should populate existing records with a script (or maybe ENS sync will do it for us) and ensure that the column is not `nullable` in the long-term.

src/modules/ens/ens.service.ts
```diff
@@ -153,10 +153,8 @@ export class EnsService {
 
     // Register event handler for new Role/App/Org
     this.ensRegistry.addListener('Transfer', async (node, owner) => {
-      const namespace = await this.domainReader.readName(node.toString());
-      if (!namespace) return;
```
jrhender, line 156: 👍

JGiter (approved):
Looks good. Also I am start thinking of introducing base Domain entity extended by Application, Organization and Role. It allows to exclude `ifOrganization`, `ifApplication` and `ifRole` checks in domains common operations like `find` or `delete`

src/modules/ens/ens.service.ts
```diff
@@ -181,15 +179,15 @@ export class EnsService {
     owner,
   }: {
     hash: string;
-    name: string;
+    name?: string;
```
jrhender, line 182:
```suggestion
```
Can we remove `name` as a param as we can read it in this event handler if we need it?

src/modules/ens/ens.service.ts
```diff
@@ -181,15 +179,15 @@ export class EnsService {
...
       const namespaceOwner = owner ? owner : await this.ensRegistry.owner(hash);
-    
+
       if (namespaceOwner === emptyAddress) {
         //prevent resync and remove namespace from database
-        return this.deleteNamespace(name);
+        return this.deleteNamespace(hash);
       }
    
       const data = await this.domainReader.read({ node: hash });
```
jrhender, line 193: Does `domainReader.read()` read return the `name`? If not, maybe we can use `domainReader.readName()` here as we now know that the namespace isn't deleted?

src/modules/ens/ens.service.unit.spec.ts
```diff
@@ -138,70 +142,75 @@ describe('EnsService', () => {
...
       jest
         .spyOn(service as any, 'getAllNamespaces')
-        .mockResolvedValueOnce(['apps.myorg.iam.ewc']);
+        .mockResolvedValueOnce(['apps.daniel.iam.ewc']);
+
       await service.syncENS();
       expect(MockLogger.error).toHaveBeenCalledWith(
         expect.stringContaining(
-          'Error syncing namespace apps.myorg.iam.ewc, owner undefined, Error: unable to parse resolved textData for node:',
+          'Error syncing namespace apps.daniel.iam.ewc, owner undefined, Error: unable to parse resolved textData for node:',
```
jrhender, line 205: A bit strange that you're changing `myorg` here for `daniel` haha. why is that?

src/modules/ens/ens.service.ts
```diff
@@ -294,4 +299,11 @@ export class EnsService {
     }
     this.logger.info('### Finished ENS Sync ###');
   }
+
+ onModuleDestroy() {
+    this.ensRegistry.removeAllListeners('Transfer');
+    this.ensRegistry.removeAllListeners('NewOwner');
+    this.domainNotifer.removeAllListeners('DomainUpdated');
+    this.publicResolver.removeAllListeners('TextChanged');
+  }
```
jrhender, line 308: This seems like a good idea, but just curious as to what triggered you to add it?
Also, not sure if this is possible, but it could be nice to add the listeners to a list when adding them and then this method could just loop through that list. In this way, if a new listener is added, then it will be automatically removed on destroy instead of needing to remember to add it. Not a necessary change though, just an idea 💡

src/modules/organization/organization.entity.ts
```diff
@@ -44,6 +44,10 @@ export class Organization implements BaseEnsEntity {
   @Column()
   namespace: string;
 
+  @Field({ nullable: true })
+  @Column({ nullable: true })
+  namehash?: string;
```
jrhender, line 49: Similar to application, I think it would be good if the namehash wan't nullable.

src/modules/organization/organization.dto.ts
```diff
@@ -70,6 +70,10 @@ export class OrganizationDTO implements BaseEnsEntity {
   @ApiProperty()
   namespace: string;
 
+  @IsString()
+  @ApiProperty()
+  namehash: string;
+
```
jrhender, line 76: I wonder if it makes sense to have the namehash be a `get` only property that we set ourselves by hashing the namespace? I'm just trying to achieve some validation that the namehash is actually correct for the namespace in question 🤔

jrhender (changes requested):
Looks good, just a few comments to address

src/migrations/1631103259350-AddNamehashColumn.ts
ahmedolaibrahim: I think it's probably best to leave it as character varying, I don't think we stand to gain much from doing so.

src/modules/application/application.entity.ts
ahmedolaibrahim: Yeah, so we need to make it nullable from the start just because of the migration for adding the column else it fails because at that point they all have it as nullable. This is only temporary though as I plan to update it after it has been populated.

The ENS sync populates the namehash column for us. I can confirm that as I tested it locally

src/modules/ens/ens.service.ts
ahmedolaibrahim: sure!

src/modules/ens/ens.service.unit.spec.ts
ahmedolaibrahim: haha. someone probably transferred  myOrg to emptyAddress so it was no longer suitable for the test case.

src/modules/ens/ens.service.ts
ahmedolaibrahim: Sometimes ago when working in the staking tests, I discovered that test were timing out due to the service still listening for events even after the tests ended. This was a fix for that issue, thought it would be a good idea to replicate here too.

src/modules/organization/organization.dto.ts
ahmedolaibrahim: This makes sense But i think we should make that as an improvement on this for later. from experience it's not straight forward implementing setters and getters with typeorm entities.

jrhender (approved):
Looks good!

PR #162, opened 2021-09-14
feat/ICS-54 adds versioning to cache server

ahmedolaibrahim:
- Adds versioning to cache server


Harasz (commented):
Why did you remove `package-lock.json`?
What about version of `DIDController`?

ahmedolaibrahim:
> Why did you remove `package-lock.json`

It's not removed, there was some changes  generated based on the `npm install` i did after pulling latest changes.

> What about version of `DIDController`?

I added versioning for DIDController. can you please check again?

Harasz (approved):

jrhender (approved):
Looks good to me

JGiter (approved):

JGiter (changes requested):
I think you also need to change url used by login strategy

jrhender:
> I think you also need to change url used by login strategy

@JGiter I think #165 has accomplished this

PR #163, opened 2021-09-15
feat(auth): change error when origins not matched

Harasz:
Change internal server error to unauthorized error when `Origin` header not matched with origin stored in JWT payload.

### DIFF
--- a/src/modules/auth/token.service.spec.ts
+++ b/src/modules/auth/token.service.spec.ts
@@ -1,4 +1,12 @@
-import { Controller, Get, INestApplication, MiddlewareConsumer, Module, Post, RequestMethod } from '@nestjs/common';
+import {
+  Controller,
+  Get,
+  INestApplication,
+  MiddlewareConsumer,
+  Module,
+  Post,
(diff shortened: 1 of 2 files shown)


jrhender:
@Harasz The functional changes look good but what's up with all of the styling changes? Were the files not confirming to the iam-cache-server styling before?

ahmedolaibrahim (commented):
Thanks @Harasz looks good. same question as @jrhender. are the style changes as a result of the pre-commit hooks we just installed?

Harasz:
@jrhender @ahmedolaibrahim previously, files did not match our prettier config, e.g.:
- Tab width should be 2, not 4 as we had previously, because we do not specify it in prettier config and the default value is 2.
- Also, print width was too long, should be 80. 
- We should have single quote, as it is specified in config.
- Also file had missing trailing commas.
We added this week a pre-commit hook to force using our styles guide.

jrhender (approved):
Looks good to me 👍

ahmedolaibrahim (approved):

PR #166, opened 2021-09-20
feat(deps): bump `iam-contracts` version to latest

Harasz:
Update `iam-contracts` package to the latest version, this step require to update `ethers`, `typechain` and `ew-did-registry` as well.


jrhender (commented):
I think it's looking good, just one comment. Let me know when the PR is ready for another review 👍

src/modules/staking/staking.service.ts
```diff
@@ -80,12 +78,9 @@ export class StakingService implements OnModuleDestroy {
     });
   }
   private InitEventListeners(): void {
-    this.stakingPoolFactory.addListener(
-      'StakingPoolLaunched',
-      async (org, address) => {
-        await this.saveStakingPool(address);
-      },
-    );
+    this.stakingPoolFactory.on('StakingPoolLaunched', async (_, address) => {
```
ahmedolaibrahim, line 81: curious question here:  is `on` equivalent to `addListener`  in the sense that it initialises and adds the listener? if not,  just thinking that we might no longer need the removeListener part.

ahmedolaibrahim (commented):
Thanks @Harasz PR looks good overall. just some few clarifying comments.

src/modules/staking/staking.service.ts
Harasz: There is no longer `addListener` method, just `on` and under the hood it is calling `_addEventListener`. So I think we still need to `removeListener` part.

ahmedolaibrahim (approved):

jrhender (approved):
Looks good!

PR #172, opened 2021-09-23
fix(ens): avoid read metadata from `roles` and `apps` domains

Harasz:
Filter out roles and apps domains in ENS event handler and avoid reading their empty metadata.

### DIFF
--- a/src/modules/ens/ens.service.ts
+++ b/src/modules/ens/ens.service.ts
@@ -194,8 +194,14 @@ export class EnsService implements OnModuleDestroy {
         return this.deleteNamespace(hash);
       }
 
-      const data = await this.domainReader.read({ node: hash });
       name = await this.domainReader.readName(hash);
+      // Don't want to sync domains that end in "roles" or "apps" (like "roles.domain.iam.ewc")
+      // This is because these are "spacer" domains that do not have metadata
+      if (name.startsWith('roles') || name.startsWith('apps')) {
+        return;
+      }
+
+      const data = await this.domainReader.read({ node: hash });
 
       if (!namespaceOwner || !data) {
         this.logger.debug(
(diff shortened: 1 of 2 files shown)


src/modules/ens/ens.service.ts
```diff
@@ -194,8 +194,13 @@ export class EnsService implements OnModuleDestroy {
         return this.deleteNamespace(hash);
       }
 
-      const data = await this.domainReader.read({ node: hash });
       name = await this.domainReader.readName(hash);
+      // Don't want to sync roles or apps domains
```
ahmedolaibrahim, line 198: is this the agreed consensus? that we don't want to sync roles or apps? why though?

ahmedolaibrahim: also you need tests for this.

Harasz: I am working on tests right now.
We do not want to sync domains that starts with `apps` and `roles` such us: `roles.domain.iam.ewc` and `apps.domain.iam.ewc`. These domains do not have `metadata` key, but we are trying to read this key in the next line 
`const data = await this.domainReader.read({ node: hash });`.

This is related to this task https://energyweb.atlassian.net/jira/software/projects/ICS/boards/86?selectedIssue=ICS-52. But we are trying to catch these domains at `getAllNamespaces` method. But we were not catch these domains in ENS listeners.

ahmedolaibrahim: okay makes senses. thanks for clarifying.

ahmedolaibrahim (approved):

src/modules/ens/ens.service.ts
jrhender:
```suggestion
      // Don't want to sync domains that end in "roles" or "apps" (like "roles.domain.iam.ewc")
      // This is because these are "spacer" domains that do not have metadata
```
I think it

jrhender (approved):
On comment on the comment 😄 , but looks good 👍

JGiter (approved):

PR #176, opened 2021-09-27
E2E auth test

Harasz:


.github/workflows/cache_server.test.yml
```diff
@@ -86,3 +86,41 @@ jobs:
           REDIS_HOST: localhost
           REDIS_PORT: ${{ job.services.redis.ports[6379] }}
           NATS_CLIENTS_URL: localhost:4222
+
+      - name: Running E2E tests
+        run: npm run test-e2e --forceExit
+        env:
+          CI: true
+          ENS_REGISTRY_ADDRESS: ${{secrets.ENS_REGISTRY_ADDRESS}}
```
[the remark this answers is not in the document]
Harasz: You mean it is better to get those values from `.en.ci` instead of Github secrets?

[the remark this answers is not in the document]
JGiter: Right. As I understand secrets are values which should not be accessible outside of CI. I don't see reason why to keep contracts private. Besides since contracts are deployed during CI we will not be able to set them ahead of tests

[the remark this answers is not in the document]
ahmedolaibrahim: @JGiter in github actions, env variable can only either be gotten from the default env variables that start with `GITHUB_` or the ones defined as `context` during execution of the workflow. since we don't want to hardcode the values during execution of the workflow, i think github secrets is right for this.

package.json
```diff
@@ -14,6 +14,7 @@
     "start:prod": "node dist/src/main",
     "lint": "eslint \"{src,apps,libs,test}/**/*.ts\" --fix",
     "test": "run-s migration:test && NODE_ENV=test jest",
+    "test-e2e": "run-s migration:test && jest --config ./test/jest-e2e.json --force-exit",
```
ahmedolaibrahim, line 17: just thinking if we really need the `jest-e2e.json` config file to achieve this as it could set a bad precedence that we have to create a config file for every test functionality type we wish to run separately, we could use regex here to select the e2e test files.

test/app.e2e-spec.ts
```diff
@@ -0,0 +1,31 @@
+import { Test } from '@nestjs/testing';
```
ahmedolaibrahim, line 1: I think the filename should follow the same convention as the other e2e tests.

Harasz: @ahmedolaibrahim  It should not. This is a test file while other files are test suite files. We have only one file with test that runs multiple test suite files. The reason for that  is to have only one instance of app running during tests because we are listening on port 3000. We could run each test one by one but it would be very time consuming.

Harasz: You want to name it `app.e2e.spec.ts`?

ahmedolaibrahim: yeah i think `auth.e2e.spec.ts` is more suitable.

PR #178, opened 2021-10-01
fix(did.service): fix parse parse logs

Harasz:
Add a new checking if the value is BigNumber during log parsing


jrhender (approved):
Looks good, thanks.

PR #179, opened 2021-10-01
test(did.service): check parsing BigNumbers

Harasz:
Test is checking if service supports current implementation of `BigNumber` from `ethers` library.


jrhender (approved):
Great, thanks 👍

PR #181, opened 2021-10-04
feat(role): add endpoint to fetch multiple roles definition

Harasz:
Add endpoint that allow fetch definition for multiple roles by one request.

### DIFF
--- a/src/modules/role/role.controller.ts
+++ b/src/modules/role/role.controller.ts
@@ -3,10 +3,12 @@ import {
   Get,
   HttpStatus,
   Param,
+  ParseArrayPipe,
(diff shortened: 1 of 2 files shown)


JGiter (approved):

jrhender:
@JGiter @Harasz I think we should not make changes to cache-server unless it is a bugfix until Oct 11 when AEMO demo is done

Harasz:
@jrhender You're right. We will merge it after AEMO demo.

ahmedolaibrahim (commented):
@Harasz looks good but i think we should have tests for this.

PR #185, opened 2021-10-11
fix(redis): remove old bull job info/metadata

hejkerooo:


jrhender:
@hejkerooo @ahmedolaibrahim I think we should wait to merge this until we get the okay from AEMO

PR #191, opened 2021-10-14
feat(claim): persists and retrieve issued tokens

Harasz:
- [x] Rename existing `Claim` entity to `RoleClaim` 
- [x] Create table `Claim`
- [x] Add endpoint for saving issued tokens
- [x] Add endpoint for retrieving issued tokens by subjects


Harasz:
Tell me guys what do you think of that. I will write some tests after your review.

PR #197, opened 2021-10-29
[ICS-85 ] DIDContacts Endpoints

ahmedolaibrahim:


jrhender:
@ahmedolaibrahim It would be better if this DID_contact functionality was separated from the DID module (into a new module). In my mind, the current DID module is about DID document (cached) resolution whereas this is really about maintaining a contact list of DIDs. What makes this need for this separation apparent to me, is I can see it being valuable down the road to split out the "DID Document resolution" functionality and the "DID address book" functionality into separate services and this will be more difficult to untangle if they are in the same module.

ahmedolaibrahim:
> @ahmedolaibrahim It would be better if this DID_contact functionality was separated from the DID module (into a new module). In my mind, the current DID module is about DID document (cached) resolution whereas this is really about maintaining a contact list of DIDs. What makes this need for this separation apparent to me, is I can see it being valuable down the road to split out the "DID Document resolution" functionality and the "DID address book" functionality into separate services and this will be more difficult to untangle if they are in the same module.

makes sense. will update.  thanks @jrhender

src/modules/did-contact/did.contact.dto.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export class verifiedRoleDTO {
+  @IsString()
+  @ApiProperty()
+  name: string;
+
+  @IsString()
+  @ApiProperty()
+  namespace: string;
+}
```
jrhender, line 32: Good idea creating these DTOs. I wonder if they would be better placed in the `auth` module though as then see to be associated with the `req.user`?

package.json
```diff
@@ -15,7 +15,7 @@
     "start:debug": "nest start --debug --watch",
     "start:prod": "node dist/src/main",
     "lint": "eslint \"{src,apps,libs,test}/**/*.ts\" --fix",
-    "test": "run-s migration:test && NODE_ENV=test jest -i \"/src/.*\\.spec\\.ts$\" --force-exit",
+    "test": "run-s migration:test && NODE_ENV=test jest -i \"/src/.*\\.spec\\.ts$\" --verbose --runInBand --force-exit",
```
jrhender, line 18: Just curious, why the `--runInBand` now?

Harasz (approved):

jrhender (commented):
Looks pretty good, in particular the tests, thanks. Just a few questions

JGiter (approved):

src/modules/did-contact/did.contact.entity.ts
```diff
@@ -0,0 +1,35 @@
...
+  PrimaryGeneratedColumn,
+} from 'typeorm';
+import { DIDDocumentEntity } from '../did/did.entity';
+
+@Entity({ name: 'did_contact' })
+export class DIDContact {
+  @PrimaryGeneratedColumn('uuid')
+  id: string;
+
+  @CreateDateColumn({ name: 'created_at', type: 'timestamp without time zone' })
```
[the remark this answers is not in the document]
ahmedolaibrahim: IMO the timezone details isn't that necessary for this feature since it isn't going to be returned to the FE as per the AC

src/modules/did-contact/did.contact.dto.ts
ahmedolaibrahim: makes sense. will update it.

package.json
ahmedolaibrahim: Mainly to aid debugging purpose. it was quite hard to debug some slow tests without it. it proved useful for me. there's also some claims that it helps speed up tests but that seems negligible in our case.

src/modules/did-contact/did.contact.entity.ts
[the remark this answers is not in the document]
jrhender: Hmm, how does this work then? Does it always save the time in UTC but without the timezone?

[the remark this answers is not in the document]
ahmedolaibrahim: yep exactly

jrhender (approved):
Looks good, thanks!

PR #199, opened 2021-11-03
[ICS-82] Disable Origin Check When ENABLE_AUTH Flag is 'false'

ahmedolaibrahim:


src/modules/auth/token.service.ts
```diff
@@ -85,6 +85,10 @@ export class TokenService {
    * A pattern such as the double cookie submit pattern cannot be used because the cache-server does not share an origin with it's clients.
    */
   async handleOriginCheck(req: Request, res: Response, next: NextFunction) {
+    const authEnabled = this.configService.get<string>('ENABLE_AUTH');
+    if (authEnabled === 'false') {
+      next();
```
Harasz, line 90:
```suggestion
      next();
      return;
```

Harasz (changes requested):

Harasz (approved):

PR #205, opened 2021-11-22
feat(claims): ICL-142 secure nats messaging in cache server

Passerino:


src/modules/claim/claim.types.ts
```diff
@@ -1,4 +1,10 @@
-export const NATS_EXCHANGE_TOPIC = 'claim.exchange';
+export const NATS_EXCHANGE_TOPIC = 'CLAIM_TOPIC';
+
+export enum ClaimRequestType {
+  CLAIM_ISSUANCE = 'CLAIM_ISSUANCE',
+  CLAIM_REJECTION = 'CLAIM_REJECTION',
+  CLAIM_ENROLLMENT_REQUEST = 'CLAIM_ENROLLMENT_REQUEST',
```
jrhender, line 6: I would simplify this. Or I would change it to `ROLE_ENROLMENT_REQUEST`. I think the word `ENROLMENT` only makes sense in the context of roles
```suggestion
  CLAIM_REQUEST = 'CLAIM_REQUEST',
```

src/modules/claim/claim.service.ts
```diff
@@ -75,73 +69,149 @@ export class ClaimService {
...
+      await this.roleService.verifyEnrolmentPrecondition({
+        claimType,
+        userDID: dto.requester,
+      });
+
+      await this.createAndIssue(dto);
+    } else if (
+      claim &&
+      !claim.isRejected &&
+      (!!rq.issuedToken || !!rq.onChainProof)
```
jrhender, line 124: Sorry, this is probably a basic Javascript question 😅 . But why the `!!` operator here? Is there a case in particular that you have in mind?

src/modules/claim/claim.controller.ts
```diff
@@ -83,14 +78,11 @@ export class ClaimController {
...
+    //const { sub } = new JWT(new Keys()).decode(data.issuedToken);
 
     const payload = JSON.stringify(claimData);
-    this.nats.connection.publish(
-      `${data.requester}.${NATS_EXCHANGE_TOPIC}`,
-      payload,
-    );
-    this.nats.connection.publish(`${sub}.${NATS_EXCHANGE_TOPIC}`, payload);
+    this.logger.debug(`publishing claims issue request ${payload} from ${did}`);
+    await this.claimProcessor.enqueueClaimIssuanceRequest(claimData);
```
jrhender, line 85: @Passerino What are the advantages/disadvantages of the queue approach in your view? I think the queue does make sense for the asynchronous portion (i.e. the messaging via NATS). However, I'm not entirely convinced that it makes sense to do the database updates via queues. I think if a database update failed for instance, I think it could be a bit odd that we returned a successful HTTP code. What are you thoughts on this?

jrhender (changes requested):
Looks like a good start, thanks for the PR! I think we need a bit of a discussion around the queue approach and just a note to remove all of the comments at some point please 😄

src/modules/claim/claim.service.ts
Passerino: I like to booleanize logical checks , but agreed, in this particular case it won't bring anything useful beside complexity. Thx for pointing out.

src/modules/claim/claim.processor.ts
```diff
@@ -1,23 +1,103 @@
...
+  public async processClaimEnrollmentRequest(job: Job<string>) {
+    const rq = JSON.parse(job.data) as IClaimRequest;
+    await this.claimService.handleClaimEnrollmentRequest(rq);
+
+    this.notifyDIDs(
+      ClaimRequestType.CLAIM_ENROLLMENT_REQUEST,
+      rq.id,
+      rq.claimIssuer,
+    );
+  }
```
jrhender, line 52: @Passerino I was thinking about this a bit more. Is `claimService.handleClaimEnrolmentRequest` idempotent? That is could we run it twice without issue?
I'm imagining a situation where `this.claimService.handleClaimEnrolmentRequest` succeeds but then `this.notifyDIDs` fails (for example because the NATS server is temporarily down). Could we run the queue job again without issue?

Passerino: whole job can be repeated according to configuration. But it anything fails , whole job will be repeated if we wish.

Passerino: If we are worried about NATS maybe we should have internal queue there in natsService...

jrhender:
```suggestion
  @Process(ClaimRequestType.CLAIM_REQUEST_NOTIFICATION_JOB)
  public async processClaimEnrollmentRequest(job: Job<string>) {
    const rq = JSON.parse(job.data) as IClaimRequest;
    await this.claimService.handleClaimEnrollmentRequest(rq); -> move this back to HTTP

    this.notifyDIDs(
      ClaimRequestType.CLAIM_ENROLLMENT_REQUEST,
      rq.id,
      rq.claimIssuer,
    );
  }
```

src/modules/claim/claim.controller.ts
```diff
@@ -136,12 +128,7 @@ export class ClaimController {
 
     const payload = JSON.stringify(claimData);
     this.logger.debug(`publishing claims request ${payload} from ${did}`);
-    data.claimIssuer.map(issuerDid => {
-      this.nats.connection.publish(
-        `${issuerDid}.${NATS_EXCHANGE_TOPIC}`,
-        payload,
-      );
-    });
+    await this.claimProcessor.enqueueClaimEnrollmentRequest(claimData);
```
jrhender, line 131:
```suggestion
    await this.claimService.handleClaimEnrollmentRequest(rq);
    if (continueWithNotification) {
      await this.claimProcessor.enqueueClaimRequestNotification(claimData);
    }
```

jrhender (commented):
Some further suggestions

src/modules/claim/claim.controller.ts
jrhender: `continueWithNotification` optional for now, no need to implement

src/modules/claim/claim.service.ts
```diff
@@ -42,20 +38,18 @@ export class ClaimService {
...
   }
 
-  private initializeExchangeListener() {
-    this.nats.connection.subscribe(`*.${NATS_EXCHANGE_TOPIC}`, async data => {
-      this.logger.debug(`Got message ${data}`);
-      await this.claimQueue.add('save', data);
-    });
-  }
+  // private initializeExchangeListener() {
+  //   this.nats.connection.subscribe(`*.${NATS_EXCHANGE_TOPIC}`, async data => {
```
hejkerooo, line 48: Please check https://github.com/energywebfoundation/iam-cache-server/pull/206
Add the same options for scaling purposes

src/modules/claim/claim.service.ts
```diff
@@ -42,20 +38,18 @@ export class ClaimService {
...
+  //     this.logger.debug(`Got message ${data}`);
+  //     await this.claimQueue.add('save', data);
+  //   });
+  // }
```
jrhender, line 52: A note to remove commented code please (prior to merge)

src/modules/claim/claim.service.ts
```diff
@@ -75,73 +69,149 @@ export class ClaimService {
...
     }
   }
 
+  /**
+   * Handles claim rejection request saving and updates.
+   * @param rq IClaimRejection request
+   */
+  public async handleClaimRejectionRequest(rq: IClaimRejection) {
+    const claim: RoleClaim = await this.getById(rq.id);
+    if (!claim || claim.isAccepted || !rq.isRejected) return;
```
jrhender, line 137: What do you think of returning some error result here so that the server can return a `400 bad request`? I think that all of these checks are for an incorrect request from the DID that made the request?

src/modules/claim/claim.service.ts
```diff
@@ -75,73 +69,149 @@ export class ClaimService {
...
+
+      await this.createAndIssue(dto);
+    } else if (
+      claim &&
+      !claim.isRejected &&
+      (!!rq.issuedToken || !!rq.onChainProof)
+    ) {
+      const dto = await ClaimIssueDTO.create(rq);
+      await this.issue(dto);
     }
```
jrhender, line 128: What do you think of returning an error here so that an error HTTP code can be returned to the client? I think it would be weird to continue to publish the message at this point is the issuance storage hasn't taken place

src/modules/nats/nats.service.ts
```diff
@@ -1,28 +1,60 @@
...
+    private natsGateway: NatsGateway,
+    private logger: Logger,
+    private readonly config: ConfigService,
+    @InjectQueue('nats-messages')
+    private readonly messagesQueue: Queue<IMessageJob>,
+  ) {
+    this.natsSubjectSuffix = this.config.get<string>('NATS_SUBJECT_SUFFIX');
+  }
+
+  public async publish(subject: string, data: Record<string, unknown>) {
```
jrhender, line 27: Is this used outside of class? Can we make it `private`?
```suggestion
  private async publish(subject: string, data: Record<string, unknown>) {
```

src/modules/nats/nats.service.ts
```diff
@@ -1,28 +1,60 @@
...
+    );
   }
 
-  async onModuleDestroy() {
-    this.connection.close();
+  @Process('message')
+  public async processMessage(job: Job<IMessageJob>) {
+    const { subject, data } = job.data;
+    this.natsGateway.connection.publish(subject, JSON.stringify(data));
+    this.natsGateway.connection.flush();
```
jrhender, line 58: Is this a best practice to `flush` after a `publish`?

jrhender (changes requested):
A bunch of comments but I think it's almost there!

src/modules/assets/assets.subscriber.ts
```diff
@@ -34,25 +34,25 @@ export class AssetsEventSubscriber {
...
-      } saved: ${JSON.stringify(saved)}`,
+      `${type} event for ${event.assetId} saved: ${JSON.stringify(saved)}`,
     );
 
     if (publish) {
-      this.nats.connection.publish(
-        `${event.relatedTo}.${NATS_EXCHANGE_TOPIC}`,
-        JSON.stringify({ ...event, type })
-      );
+      this.nats.publish(`${event.relatedTo}.${NATS_EXCHANGE_TOPIC}`, {
```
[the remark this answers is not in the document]
Passerino: +1

docs/api/modules/modules_claim_claim_types.md
```diff
@@ -41,4 +42,4 @@
 
 ### NATS\_EXCHANGE\_TOPIC
 
-• **NATS\_EXCHANGE\_TOPIC**: ``"claim.exchange"``
+• **NATS\_EXCHANGE\_TOPIC**: ``"claim-topic"``
```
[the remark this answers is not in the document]
Passerino: +1

src/modules/claim/claim.module.ts
```diff
@@ -1,27 +1,23 @@
-import { BullModule } from '@nestjs/bull';
+//import { BullModule } from '@nestjs/bull';
```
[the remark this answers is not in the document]
Passerino: +1

src/modules/nats/nats.service.ts
Passerino: Jakub and Kris investigated the problem. It seems that without flush there is a risk of not sending the message at all.

src/modules/nats/nats.service.ts
Passerino: It's in use in one place , I'll confirm if we can turn it into publishDids and hide this one.

src/modules/nats/nats.service.ts
jrhender: @Harasz @hejkerooo when you have a moment, could you comment with any additional information you have about why a `flush` after `publish` is necessary?

Harasz: https://docs.nats.io/using-nats/developer/sending/caches

> For performance reasons, most if not all, of the client libraries will buffer outgoing data so that bigger chunks can be written to the network at one time. This may be as simple as a byte buffer that stores a few messages before being pushed to the network.

@jrhender

src/modules/claim/claim.controller.ts
```diff
@@ -134,14 +140,20 @@ export class ClaimController {
...
 
-    const payload = JSON.stringify(claimData);
-    this.logger.debug(`publishing claims request ${payload} from ${did}`);
-    data.claimIssuer.map(issuerDid => {
-      this.nats.connection.publish(
-        `${issuerDid}.${NATS_EXCHANGE_TOPIC}`,
-        payload,
-      );
-    });
+    const result = await this.claimService.handleClaimEnrollmentRequest(
```
jrhender, line 143: Minor point, but please rename `handleClaimEnrollmentRequest` to `handleClaimEnrolmentRequest` as this matches the naming convention used across the IAM stack
```suggestion
    const result = await this.claimService.handleClaimEnrolmentRequest(
```

src/modules/claim/claim.service.ts
```diff
@@ -33,6 +29,18 @@ interface QueryFilters {
 
 export const UUID_NAMESPACE = '5193850c-2367-4ec4-8c22-95dfbd4a2880';
 
+export class ClaimHandleResult {
+  isSuccessful: boolean;
+  deatils?: string;
```
jrhender, line 34: Minor spelling fix
```suggestion
  details?: string;
```

src/modules/claim/claim.service.ts
```diff
@@ -81,71 +73,91 @@ export class ClaimService {
   }
 
   /**
-   * Handles claims saving and updates
-   * @param data Raw claim data
+   * Handles claim enrollment request saving and updates.
```
jrhender, line 76:
```suggestion
   * Handles claim enrolment request saving and updates.
```

src/modules/claim/claim.service.ts
```diff
@@ -81,71 +73,91 @@ export class ClaimService {
...
+   * @param rq IClaimRequest request
+   */
+  public async handleClaimEnrollmentRequest(
```
jrhender, line 79:
```suggestion
  public async handleClaimEnrolmentRequest(
```

src/modules/claim/claim.service.ts
```diff
@@ -81,71 +73,91 @@ export class ClaimService {
...
+    rq: IClaimRequest,
+  ): Promise<ClaimHandleResult> {
+    const claim: RoleClaim = await this.getById(rq.id);
+    if (claim || !rq.token)
+      return ClaimHandleResult.Failure(
+        'Claim Enrollment Request criteria not met',
```
jrhender, line 85:
```suggestion
        'Claim Enrolment Request criteria not met',
```

src/modules/claim/claim.service.ts
```diff
@@ -81,71 +73,91 @@ export class ClaimService {
...
+      await this.createAndIssue(dto);
+    } else if (
+      claim &&
+      !claim.isRejected &&
+      (rq.issuedToken || rq.onChainProof)
+    ) {
+      const dto = await ClaimIssueDTO.create(rq);
+      await this.issue(dto);
+    } else {
+      return ClaimHandleResult.Failure('Non of claim issuance criteria met');
```
jrhender, line 142:
```suggestion
      return ClaimHandleResult.Failure('claim issuance criteria not met');
```

src/modules/claim/claims.controller.e2e.spec.ts
```diff
@@ -156,11 +144,11 @@ describe('ClaimsController', () => {
   });
 
   afterEach(async () => {
-    await queryRunner.rollbackTransaction();
+    await queryRunner?.rollbackTransaction();
     await app.close();
   });
 
-  it('getBySubject() should return claim requested for subject', async () => {
+  it.only('getBySubject() should return claim requested for subject', async () => {
```
jrhender, line 151:
```suggestion
  it('getBySubject() should return claim requested for subject', async () => {
```

src/modules/claim/claim.types.ts
```diff
@@ -1,4 +1,10 @@
-export const NATS_EXCHANGE_TOPIC = 'claim.exchange';
+export const NATS_EXCHANGE_TOPIC = 'claim-exchange';
+
+export enum ClaimRequestType {
+  CLAIM_ISSUANCE = 'credential-issued',
+  CLAIM_REJECTION = 'credential-rejected',
+  CLAIM_ENROLMENT_REQUEST = 'request-credential',
+}
```
jrhender, line 7: I suggest we align the names of the event names from Aries protocol to remind us to stay in line with the specifications where we can 😄 
```suggestion
/**
 * Types inspired by Aries Issue-Credential protocol {@link https://github.com/hyperledger/aries-rfcs/blob/main/features/0453-issue-credential-v2/README.md}
 */
export enum ClaimRequestType {
  CLAIM_ISSUANCE = 'issue-credential',
  CLAIM_REJECTION = 'reject-credential',
  CLAIM_REQUEST = 'request-credential',
}
```

src/modules/claim/claim.types.ts
```diff
@@ -1,4 +1,10 @@
-export const NATS_EXCHANGE_TOPIC = 'claim.exchange';
+export const NATS_EXCHANGE_TOPIC = 'claim-exchange';
```
jrhender, line 1: What's the relationship between `NATS_EXCHANGE_TOPIC` and `NATS_SUBJECT_SUFFIX` from the `.env`?

jrhender (changes requested):
Some suggestions + need to remove `it.only`

src/modules/nats/nats.service.ts
jrhender: @Harasz Hmm, interesting, thanks. You'd think NATS would send the data as a small chunk after a certain amount if it wasn't able to assemble a bigger chunk. So this is kind-of a workaround past a NATS performance optimization 🤔 ?

Harasz: Using flush after publish in this situation we are publishing event immediately. We will not wait until the buffer is full. This results in disabling default NATS optimization, but also event about claims will be published immediately.

jrhender (approved):
Looks good! Let's merge!

PR #206, opened 2021-11-25
feat(nats): use queues for processing nats messages

hejkerooo:
Motivation: https://docs.nats.io/using-nats/developer/receiving/queues

### DIFF
--- a/src/modules/claim/claim.service.ts
+++ b/src/modules/claim/claim.service.ts
@@ -51,10 +51,16 @@ export class ClaimService {
   }
 
   private initializeExchangeListener() {
-    this.nats.connection.subscribe(`*.${NATS_EXCHANGE_TOPIC}`, async data => {
-      this.logger.debug(`Got message ${data}`);
-      await this.claimQueue.add('save', data);
-    });
+    this.nats.connection.subscribe(
+      `*.${NATS_EXCHANGE_TOPIC}`,
+      {
+        queue: 'iam_cache_server_exchange_topic',
+      },
+      async data => {
+        this.logger.debug(`Got message ${data}`);
+        await this.claimQueue.add('save', data);
+      },
+    );
   }
 
   /**


Harasz (approved):

src/modules/claim/claim.service.ts
```diff
@@ -51,10 +51,16 @@ export class ClaimService {
...
 
   private initializeExchangeListener() {
-    this.nats.connection.subscribe(`*.${NATS_EXCHANGE_TOPIC}`, async data => {
-      this.logger.debug(`Got message ${data}`);
-      await this.claimQueue.add('save', data);
-    });
+    this.nats.connection.subscribe(
+      `*.${NATS_EXCHANGE_TOPIC}`,
+      {
+        queue: 'iam_cache_server_exchange_topic',
```
jrhender, line 57: @hejkerooo where is this queue name from? Is this configured in NATS?

src/modules/claim/claim.service.ts
```diff
@@ -51,10 +51,16 @@ export class ClaimService {
...
+      },
+      async data => {
+        this.logger.debug(`Got message ${data}`);
+        await this.claimQueue.add('save', data);
+      },
+    );
```
jrhender, line 63: @Passerino with our messaging changes, I don't think we expect the cache-server to be receiving any messages at this moment, right? All communication **to** the cache-server is done via HTTP, I believe. Perhaps we can remove this code?

jrhender (commented):
Some additional questions

src/modules/claim/claim.service.ts
jrhender: @hejkerooo, thanks for the change btw, I didn't know about NATS queues, this is pretty awesome

src/modules/claim/claim.service.ts
jrhender: @Passerino  Ah, nevermind, I see you have commented it out in your PR already, so I think we can remove it in your PR 😄  https://github.com/energywebfoundation/iam-cache-server/pull/205/files

PR #208, opened 2021-11-29
feat: migrating existing did to have chain id

Harasz:


src/migrations/1638185287373-AddChainIDToDIDs.ts
```diff
@@ -0,0 +1,91 @@
...
+    await queryRunner.query(this.getUpUpdateSQL('asset', `"documentId"`));
+
+    // assets_history table
+    await queryRunner.query(
+      this.getUpUpdateSQL('assets_history', `"emittedBy"`),
+    );
+    await queryRunner.query(
+      this.getUpUpdateSQL('assets_history', `"relatedTo"`),
+    );
+    await queryRunner.query(this.getUpUpdateSQL('assets_history', `"assetId"`));
```
jrhender, line 39: The asset contracts aren't really aware of DIDs at the moment. These tables storing DIDs currently? I guess they are 🤔 ?

jrhender (commented):
Some initial comments

src/migrations/1638185287373-AddChainIDToDIDs.ts
Harasz: I got the backup of the dev database. I search for all DIDs there and what I found is here.

jrhender: Hmm, ok, let's keep the DIDs for the asset column for now 👍

src/modules/claim/claims.controller.e2e.spec.ts
```diff
@@ -200,14 +204,14 @@ describe('ClaimsController', () => {
         expect(res.body.length).toEqual(1);
         expect(res.body[0]).toBeInstanceOf(Object);
         expect(res.body[0].issuedToken).toEqual(token);
-        expect(res.body[0].issuedAt).toHaveLength(10);
+        expect(res.body[0].issuedAt).not.toBeNaN();
```
jrhender, line 207: This seems like it makes the test less specific. Could we keep both conditions? Or is the `NaN` condition more appropriate?

src/modules/claim/claims.controller.e2e.spec.ts
```diff
@@ -200,14 +204,14 @@ describe('ClaimsController', () => {
...
         expect(res.body[0].subject).toEqual(requester);
       });
   });
 
   it('`/user/:did` should return only claim related to authenticated user', async () => {
-    const requester = Wallet.createRandom().address;
-    const foreignRequester = Wallet.createRandom().address;
+    const requester = randomDID();
+    const foreignRequester = randomDID();
```
jrhender, line 214: Weird that this test works still because it was just `address` before

jrhender (commented):
Some questions. Will continue PR review soon

src/modules/claim/claims.controller.e2e.spec.ts
Harasz: I think it is more specific. `issuedAt` should be a timestamp, so nor we are checking if this is a number, but previously we were just checking for any string with length 10.

src/modules/claim/claims.controller.e2e.spec.ts
Harasz: We do not have validation, so we can pass any string there and it will pass.

src/modules/claim/claims.controller.e2e.spec.ts
jrhender: Hmm, I think `NaN` is a type in javascript, right? So this test is saying that `issuedAt` just has to be not `NaN`. But most things are not `NaN`, e.g. a `number`, a `string`, an `object`, an `array`, etc. So I'm not seeing how this is more specific. Could you explain a bit more please?

src/modules/claim/claims.controller.e2e.spec.ts
jrhender: Ah okay, I see, thanks

src/modules/claim/entities/roleClaim.entity.ts
```diff
@@ -1,18 +1,18 @@
...
+import { IRoleClaim, RegistrationTypes } from '../claim.types';
 
 @ObjectType()
 @Entity({ name: 'role_claim' })
 export class RoleClaim implements IRoleClaim {
   static create(data: Partial<RoleClaim>): RoleClaim {
     const entity = new RoleClaim();
     const jwt = new JWT(new Keys());
-    data.subject = jwt.decode(data.token).sub as string;
+    data.subject = (jwt.decode(data.token) as { sub: string }).sub;
```
jrhender, line 15: I think this change is fine, but just curious why it is useful. Is it perhaps easier to read in your view? Or is it safer somehow?

src/modules/did/did.service.ts
```diff
@@ -198,7 +198,7 @@ export class DIDService {
   private async InitEventListeners(): Promise<void> {
     const DIDAttributeChanged = 'DIDAttributeChanged';
     this.didRegistry.on(DIDAttributeChanged, async address => {
-      const did = `did:${Methods.Erc1056}:${address}`;
+      const did = `did:${Methods.Erc1056}:${process.env.CHAIN_NAME}:${address}`;
```
jrhender, line 201: 👍

test/auth/login.testSuite.ts
```diff
@@ -26,7 +26,7 @@ export const authLoginTestSuite = () => {
         .expect(401);
     });
 
-    it(`should login user`, async () => {
+    it.only(`should login user`, async () => {
```
jrhender, line 29:
```suggestion
    it(`should login user`, async () => {
```

src/modules/claim/claims.controller.e2e.spec.ts
Harasz: I am checking if `issuedAt` is not a not  number, so actually I can pass whatever type i want to NaN and checking if I could parse it to number later on. Since we have double negation we are checking if it can be parse to number

src/modules/did/did.types.ts
```diff
@@ -2,24 +2,49 @@
...
+      didMethod = idParts[1];
+      didChain = undefined;
+      didId = idParts[2];
+    } else if (idParts.length === 4) {
+      didMethod = idParts[1];
+      didChain = idParts[2];
+      didId = idParts[3];
+    } else {
+      throw new Error('Unsupported DID format');
+    }
```
jrhender, line 37: I think this code looks good but wondering if you've seen similar code elsewhere? I wonder if this logic could be encapsulated in ew-did-registry as there is probably similar logic there, I think 🤔

jrhender (changes requested):
Some additional comments + need to remove an `only` in a test

src/modules/claim/entities/roleClaim.entity.ts
Harasz: Basically I updated did-registry and it new version it appear that `jwt.decode` return `unknown`, so Typescript start to mark it as error.

src/modules/did/did.types.ts
Harasz: I copied it from `did-registry` since there was no export for DID class

jrhender: Ok, I will export now so that you can use, I think that is better than introducing this duplication

src/modules/claim/entities/roleClaim.entity.ts
jrhender: Ok, interesting. Fair enough 👍

src/modules/did/did.types.ts
jrhender: Mmm, actually, let's do that afterwards. Let's get this merged for now

jrhender (approved):
Looks good 👍

jrhender (changes requested):
Actually, let me just look at the failing tests first

src/migrations/1638185287373-AddChainIDToDIDs.ts
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  private getDownUpdateSQL(table: string, column: string): string {
+    return `UPDATE public.${table} SET ${column} = split_part(${column},':', 1) || ':' || split_part(${column},':', 2) || ':' || split_part(${column},':', 4) WHERE ${column} IS NOT NULL;`;
+  }
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    this.checkForChainID();
+
+    await queryRunner.query("SET session_replication_role = 'replica';");
```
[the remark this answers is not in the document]
Harasz: yes, that's right

src/modules/auth/token.service.ts
```diff
@@ -92,6 +92,10 @@ export class TokenService {
     let token = null;
     if (req.headers['authorization']) {
       token = req.headers['authorization'].replace('Bearer ', '');
+      console.log(
+        req.headers['authorization'],
+        req.headers['authorization'].replace('Bearer ', ''),
```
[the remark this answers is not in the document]
Harasz: Yeah, already removed

.env.ci
```diff
@@ -17,6 +17,7 @@ NATS_ENVIRONMENT_NAME=$NATS_ENVIRONMENT_NAME
 
 # ENS
 CHAIN_ID=73799
+CHAIN_NAME=volta
```
jrhender, line 20: I think the chain info is relevant to more than just ENS (i.e. it's obviously relevant to DID caching as well), so I think I would split it out into another section
```suggestion
# CHAIN CONNECTION
CHAIN_ID=73799
CHAIN_NAME=volta

# ENS
```

jrhender (commented):
Comment about env var that I noticed

package.json
```diff
@@ -91,7 +91,8 @@
     "helmet": "^4.4.1",
     "lodash.chunk": "^4.2.0",
     "nats": "^1.4.12",
-    "passport-did-auth": "^1.0.0-alpha.23",
+    "passport": "^0.4.1",
```
jrhender, line 94: @Harasz Do you know why `passport` was explicitly specified here?

jrhender (approved):
Looks good. I think failing build is some commit ordering issue but shouldn't be a problem.

PR #215, opened 2021-12-03
feat(nats): removed unnecessary fields from messages

Passerino:
url field not in use anymore due to the fact we are fetchin claims through iam-client-lib now


Harasz (approved):

jrhender (approved):
Looks good

PR #217, opened 2021-12-06
Task/ics 102 migrate did jh changes

jrhender:
- Updating packages to fix issues with tests and revert to a v1 lockfile
- Make DID caching compatible with old DIDs


JGiter (approved):
Looks good. Though if valid DID's should have chain property it would be better to ensure that DID constructor returns such did's.

PR #220, opened 2021-12-07
fix/ICS-117 assets event type rollback

Passerino:
asset event type back to uppercase due to too many dependencies


jrhender (approved):
Looks good 👍 . Maybe we can align on lowercase on day in the future 😄

PR #221, opened 2021-12-07
fix(claims) issue claim on-chain without issuedToken

Passerino:
on chain claim issuance does not need issuedToken present


jrhender (approved):
👍

PR #224, opened 2021-12-07
test(claim): add claim issuance e2e tests

Harasz:


src/modules/claim/claim.service.ts
```diff
@@ -112,12 +112,16 @@ export class ClaimService {
     rq: IClaimIssuance,
   ): Promise<ClaimHandleResult> {
     const claim: RoleClaim = await this.getById(rq.id);
+    const {
+      claimData: { claimType, claimTypeVersion },
+    } = jwt.decode(rq.issuedToken) as DecodedClaimToken;
```
jrhender, line 117: @Harasz This needs to be moved into first `if`. See https://github.com/energywebfoundation/iam-cache-server/pull/221/files

JGiter (approved):

jrhender (changes requested):
just that one previously suggested change is necessary to avoid a regression of https://energyweb.atlassian.net/browse/ICS-117
Other than that, it looks great. Really appreciate the addition of the tests 👍 !

jrhender (approved):
Looks good, thanks

PR #226, opened 2021-12-08
feat(logs): enable logs in console to display in argo

Passerino:
logs are now being sent to file and console on every env

### DIFF
--- a/src/modules/logger/logger.service.ts
+++ b/src/modules/logger/logger.service.ts
@@ -24,25 +24,6 @@ export class Logger extends NestLogger implements LoggerService {
     private readonly sentryService: SentryService,
   ) {
     super();
-    const isProduction = configService.get<string>('NODE_ENV') === 'production';
-    const logFormat = isProduction
-      ? format.combine(format.timestamp(), format.json())
-      : format.combine(
-          format.timestamp(),
-          format.printf(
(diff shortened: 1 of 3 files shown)


src/modules/logger/logger.service.ts
```diff
@@ -54,14 +35,31 @@ export class Logger extends NestLogger implements LoggerService {
...
+      maxFiles: '14d',
+    });
+
+    const console = new transports.Console({ format: logFormat });
 
     this.logger = createLogger({
       format: logFormat,
       level: 'debug',
-      transports: [transport],
+      transports: [console, file],
```
jrhender, line 62: @Passerino this means that we are logging to both `console` and `file` then?

Passerino: yes, I don't think we will look at the file log anytime soon, but wanted to keep backward compatibility.

jrhender: @hejkerooo do you see any issue with logging to a file in a kubernetes pod?

jrhender (approved):
Let's try it out 😄  If there are issues, we can revert

src/modules/logger/logger.service.ts
Passerino: it's a rolling file so I don't foreseen any storage limit issues.

PR #227, opened 2021-12-08
bug/ICS-122 did not backwards compatible

JGiter:
- Pipe which transforms DID to appropriate form

### DIFF
--- a/src/modules/did/did.types.ts
+++ b/src/modules/did/did.types.ts
@@ -1,11 +1,17 @@
-import { Chain, Methods } from '@ew-did-registry/did';
+import { Methods } from '@ew-did-registry/did';
+import { MalformedDIDError } from '../../errors/MalformedDIDErrro';
 
+// did-reg Chain is lowercase
+export enum Chain {
+  VOLTA = 'VOLTA',
+  EWC = 'EWC',
+}
 /**
  * @todo >> ew-did-registry/did
  * matches did:ethr:volta:address and did:ethr:address
  */
 export const didPattern =
-  '^(?:did:(?<method>[a-z0-9]+?):)(?<chain>[a-z0-9]+?:)?(?<id>0x[A-Fa-f0-9]{40})$';
+  '^(?:did:(?<method>[a-z0-9]+?):)((?<chain>[a-zA-Z0-9]+?):)?(?<id>0x[A-Fa-f0-9]{40})$';
 
 /**
  * A value object representing EIP-1056 DID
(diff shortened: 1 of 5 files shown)


jrhender (commented):
This looks like a great start! Good call on the pipe 👍 

Just need to set the `chain` in `DID` to `CHAIN_NAME` from config if it is missing, I think 👍

src/modules/did/did.types.ts
```diff
@@ -7,39 +17,33 @@ export class DID {
...
+    this.chain = <Chain>chain;
+    this.id = id;
+    // If an ethr DID is received that doesn't have chain info, then add it.
+    // This is because we want to be backwards compatible with EWF code that didn't include the chain info in the DID
+    // But we want to add the chain info because the DID resolution is actually occurring from a specific chain
+    if (this.chain) {
+      this.did = `did:${this.method}:${this.chain}:${this.id}`;
     } else {
-      throw new Error('Unsupported DID format');
+      this.did = `did:${this.method}:$${this.id}`;
```
Harasz, line 45:
```suggestion
      this.did = `did:${this.method}:${this.id}`;
```
??

Harasz (changes requested):
What about other places? I think we are querying DID in a couple of other places. In the latest migration file (1638185287373-AddChainIDToDIDs.ts) you can find all places in database where we are using DIDs.

JGiter:
> What about other places? I think we are querying DID in a couple of other places. In the latest migration file (1638185287373-AddChainIDToDIDs.ts) you can find all places in database where we are using DIDs.

It requires more efforts to make all endpoints backward compatible. For example piping `did` query parameter of claims endpoints requires piping accompanying `IClaimIssuance`, `IClaimRequest` and `IClaimRejection`. I think it is risky to start such big changes during release @jrhender ?

jrhender (approved):
This looks good.

jrhender:
> It requires more efforts to make all endpoints backward compatible. For example piping did query parameter of claims endpoints requires piping accompanying IClaimIssuance, IClaimRequest and IClaimRejection. I think it is risky to start such big changes during release @jrhender ?

Yes, I think it's better to get this changed merged and then we make further changes if we need to.

jrhender:
Looking at `1638185287373-AddChainIDToDIDs.ts`, I think these are all the spots:
```
    // asset table
    await queryRunner.query(this.getUpUpdateSQL('asset', 'id'));
    await queryRunner.query(this.getUpUpdateSQL('asset', 'owner'));
    await queryRunner.query(this.getUpUpdateSQL('asset', `"offeredTo"`));
    await queryRunner.query(this.getUpUpdateSQL('asset', `"documentId"`));

    // assets_history table
    await queryRunner.query(
      this.getUpUpdateSQL('assets_history', `"emittedBy"`),
    );
    await queryRunner.query(
      this.getUpUpdateSQL('assets_history', `"relatedTo"`),
    );
    await queryRunner.query(this.getUpUpdateSQL('assets_history', `"assetId"`));

    // did_contact table
    await queryRunner.query(this.getUpUpdateSQL('did_contact', 'did'));
    await queryRunner.query(this.getUpUpdateSQL('did_contact', 'created_by'));

    // did_document_entity table
    await queryRunner.query(this.getUpUpdateSQL('did_document_entity', 'id'));

    // role_claim table
    await queryRunner.query(this.getUpUpdateSQL('role_claim', 'requester'));
    await queryRunner.query(this.getUpUpdateSQL('role_claim', 'subject'));
    await queryRunner.query(this.getUpUpdateSQL('role_claim', `"acceptedBy"`));
```
I think we're still not backwards compatible on the `did_contact` and `role_claim`. But I think folks are accessing those endpoints through Switchboard, which has the latest iam-client-lib, so I think we're okay to leave those 🤞

JGiter:
Ok. Will deal with rest endpoints later

jrhender:
@JGiter let's do it in a separate story if necessary, IMO. Your talents can be put to use elsewhere for now.
If @Harasz is okay with it, let's merge this one

Harasz (approved):
Fair enough, let's merge it

PR #234, opened 2021-12-09
fix: update claimmanager address

nichonien:
Update ClaimManager address to the new deployed upgradable UUPS contract.


jrhender:
@nichonien let's just delete the CLAIM_MANAGER env variable until we need it IMO https://energywebfoundation.slack.com/archives/C028ZC4G2RE/p1638982881196000

JGiter (approved):
Probably it should also be updated in iam-client-lib and switchboard

PR #238, opened 2021-12-13
feat: add typeorm logging

hejkerooo:


src/db/config.ts
```diff
@@ -17,7 +17,7 @@ export const getDBConfig = (configService: ConfigService) => {
     cli: { migrationsDir: 'src/migrations' },
     migrationsRun: true,
     migrationsTableName: 'migrations_iam_cache_server',
-    logging: false,
+    logging: ['error', 'migration', 'warn', 'info'],
```
jrhender, line 20: It would be better to have this as config so that we can control per environment
```suggestion
    logging: JSON.parse(configService.get<string>('TYPEORM_LOGGING')),
```

jrhender (changes requested):

jrhender (approved):
Looks good

PR #242, opened 2021-12-14
feat(logs): partial redaction of DIDs

Passerino:
partial redaction of DIDs allows trackin while keeping security level high


jrhender (approved):
Looks good, thanks

PR #245, opened 2021-12-14
Ics 131 verify issuer signature

nichonien:
This PR consists of changes that verifies the `issuedToken` and `onChainProof` against the issuer DID or it's delegates key.


src/modules/claim/claim.controller.ts
```diff
@@ -75,6 +78,16 @@ export class ClaimController {
     @Param('did') did: string,
     @Body() data: IClaimIssuance,
   ) {
+    const jwt = new JWT(new Keys());
+    const didDoc = await this.didService.getById(did);
+    const proofVerifier = new ProofVerifier(didDoc);
+    if (!(await proofVerifier.verifyAssertionProof(data.issuedToken))) {
```
JGiter, line 84: This will allow fake `onChainProof` when `issuedToken` is authentic

JGiter (changes requested):
I think `issue with role issuer type ROLE` test fails because issuer document is empty https://github.com/energywebfoundation/iam-cache-server/blob/3a170d2a3b13a4f2af693996b1da1b96da16f1f3/test/claim/index.ts#L221. Document should contain id and service endpoint with claimType `role.roles.iam.ewc`

JGiter (approved):

src/modules/claim/claim.controller.ts
```diff
@@ -75,14 +79,30 @@ export class ClaimController {
...
+    ) {
+      throw new ForbiddenException('User signature not valid');
+    }
+
+    if (
+      data.onChainProof &&
+      !(await proofVerifier.verifyAssertionProof(data.onChainProof))
+    ) {
+      throw new ForbiddenException('User signature not valid');
+    }
```
jrhender, line 97: @nichonien @JGiter I think this is incorrect. I don't think that `proofVerifier.verifyAssertionProof` can verify `onChainProof`, right?

jrhender (commented):
Comment with possible issue

src/modules/claim/claim.controller.ts
JGiter: Ah, right. It is only signature part of jwt expected by `ProofVerifier`. So this check probably can be removed because identity of issuer of onchain proof is verified by Claim Manager

PR #246, opened 2021-12-14
feature/ICS-134 persist rejection reason

JGiter:
https://energyweb.atlassian.net/browse/ICS-134

### DIFF
--- a/src/modules/claim/claim.dto.ts
+++ b/src/modules/claim/claim.dto.ts
@@ -157,6 +157,10 @@ export class ClaimRejectionDTO implements IClaimRejection {
   @IsBoolean()
   @ApiProperty()
   isRejected: boolean;
+
+  @IsString()
+  @IsOptional()
+  rejectionReason?: string;
 }
 
 export class IssuedClaimDTO {
--- a/src/modules/claim/claim.types.ts
+++ b/src/modules/claim/claim.types.ts
@@ -30,6 +30,7 @@ export interface IClaimIssuance extends IMessage {
 
 export interface IClaimRejection extends IMessage {
   isRejected: boolean;
+  rejectionReason?: string;
 }
 
 export type DecodedClaimToken = {
@@ -54,6 +55,7 @@ export interface IRoleClaim {
   isAccepted: boolean;
   acceptedBy?: string;
   isRejected?: boolean;
+  rejectionReason?: string;
   namespace: string;
 }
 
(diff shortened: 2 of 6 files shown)


src/modules/claim/claim.types.ts
```diff
@@ -41,6 +41,13 @@ export enum RegistrationTypes {
   OnChain = 'RegistrationTypes::OnChain',
 }
 
+export enum RejectionReason {
+  ALREADY_ISSUED = 'Email already associated with DID',
+  MAIL_IS_BLACKLISTED = 'Email belongs to blacklisted domain',
+  CONFIRMATION_LINK_EXPIRED = 'The email confirmation link has expired',
+  VERIFICATION_SEND_ERROR = 'Unable to send verification email to provided email address',
+}
+
```
jrhender, line 50: I don't think this is necessary. The issuer/rejector should be able to put whatever they want
```suggestion
```

src/modules/claim/claim.dto.ts
```diff
@@ -157,6 +158,10 @@ export class ClaimRejectionDTO implements IClaimRejection {
   @IsBoolean()
   @ApiProperty()
   isRejected: boolean;
+
+  @IsEnum(RejectionReason)
```
jrhender, line 162: Let's keep it simple. Maybe a max-length though would be good though. Maybe 200 char or something?
```suggestion
  @IsString()
```

src/modules/claim/claim.types.ts
```diff
@@ -54,6 +61,7 @@ export interface IRoleClaim {
   isAccepted: boolean;
   acceptedBy?: string;
   isRejected?: boolean;
+  rejectionReason?: RejectionReason;
```
jrhender, line 64:
```suggestion
  rejectionReason?: string;
```

src/modules/claim/entities/roleClaim.entity.ts
```diff
@@ -74,6 +74,10 @@ export class RoleClaim implements IRoleClaim {
   @Column({ nullable: true, type: 'bool', default: false })
   isRejected?: boolean | null;
 
+  @Field({ nullable: true })
+  @Column({ nullable: true, type: 'enum', enum: RejectionReason })
+  rejectionReason?: RejectionReason;
```
jrhender, line 79: change to string

jrhender (changes requested):
I think we should keep it simple and let the requestor/issuer define the rejection reason. Not only does it simplify the code, it isn't realistic to have to add values to the `enum` for every possible reason for every possible use case

jrhender:
@JGiter why are there 295 files changed? Is this expected?

JGiter:
> @JGiter why are there 295 files changed? Is this expected?

Probably these .md left from previous build. Cleared them

jrhender (approved):
Looks good to me, thanks @JGiter 😄

PR #249, opened 2021-12-16
bug/ICS-141 access tokens are sent to sentry

JGiter:
https://energyweb.atlassian.net/browse/ICS-141
- [x] strip access and refresh tokens from request cookies and headers
- [x]  strip refresh token from request query
- [x] remove identityToken from request body

### DIFF
--- a/src/modules/sentry/sentry.service.ts
+++ b/src/modules/sentry/sentry.service.ts
@@ -41,12 +41,14 @@ export class SentryService {
         const request = event.request;
         if (request) {
           delete request.cookies;
+          delete request.data.identityToken;
+          request.url.replace(/refresh_token=.*/, 'refresh_token=[Filtered]');
           const headers = request.headers;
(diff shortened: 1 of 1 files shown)


jrhender (commented):
Looks like a good start to me 👍

src/modules/sentry/sentry.service.ts
```diff
@@ -37,6 +37,20 @@ export class SentryService {
         }),
         new Sentry.Integrations.OnUnhandledRejection({ mode: 'warn' }),
       ],
+      beforeSend(event) {
+        const request = event.request;
+        if (request) {
+          delete request.cookies;
+          delete request.data.identityToken;
```
jrhender, line 44: Maybe safe index would be better?
```suggestion
          delete request.data?.identityToken;
```

Harasz (approved):

PR #250, opened 2021-12-16
fix(claims): ensure only one event is propagated to each notified DID

Passerino:
remove duplicated DIDs , fixed problem with wrong subject in case of issuedToken not present


jrhender (approved):
Seems good to me, thanks @Passerino

PR #256, opened 2021-12-27
feat(claim): remove unused DID param in claim request path

Harasz:


src/modules/claim/claim.controller.ts
```diff
@@ -164,7 +161,7 @@ export class ClaimController {
       { claimId: claimData.id },
     );
 
-    this.logger.debug(`credentials requested from ${did}`);
+    this.logger.debug(`credentials requested from ${sub}`);
```
JGiter, line 164:
```suggestion
    this.logger.debug(`credentials requested from ${requester} for ${sub}`);
```

JGiter (changes requested):

JGiter (approved):

PR #257, opened 2021-12-28
feat(auth): add endpoint for checking auth status

Harasz:


src/modules/auth/login.controller.ts
```diff
@@ -132,4 +132,28 @@ export class LoginController {
 
     return res.send({ token, refreshToken });
   }
+
+  @Get('auth/status')
+  async status(@Req() req: Request) {
```
JGiter, line 137: This check is supposed to ensure that access token is matching to requester DID instead of to DID from previous session. I guess it should just return user extracted from `@User` decorator

JGiter (changes requested):

src/modules/auth/login.controller.ts
Harasz: I will update it

src/modules/auth/login.controller.ts
```diff
@@ -132,4 +132,28 @@ export class LoginController {
...
+    const accessTokenString =
+      req.cookies[this.configService.get<string>('JWT_ACCESS_TOKEN_NAME')] ||
```
Passerino, line 139: not a big deal but shouldn't header be primary source of truth?

Harasz: I think that depends from the environment, but token is first trying to be extracted from header in the login strategy.

JGiter (approved):
Although this should work I am curios if we can reuse JwtStrategy service. For example by invoking JwtStrategy.authenticate(req) and then searching for `user` property in request

test/auth/authStatus.testSuite.ts
```diff
@@ -0,0 +1,72 @@
...
+
+    beforeAll(async () => {
+      provider = new providers.JsonRpcProvider(process.env.ENS_URL);
+      wallet = new Wallet(
+        '779907598c747ff45a4f8e1b7e0fde0756585a9f936aecc95c1c738a3d85bbc4',
+        provider,
+      );
+    });
+
+    it(`should not throw error if used is not logged in`, () => {
```
jrhender, line 19: small typo
```suggestion
    it(`should not throw error if user is not logged in`, () => {
```

src/modules/auth/login.controller.ts
```diff
@@ -132,4 +132,31 @@ export class LoginController {
...
+      );
+
+      return {
+        user: tokenData?.did || null,
+      };
+    } catch {
+      return {
+        user: null,
+      };
+    }
```
jrhender, line 160: Why the `try/catch` here? Are we expecting errors from `verifyAccessToken` in some cases?
```suggestion
      const tokenData = await this.tokenService.verifyAccessToken(
        accessTokenString,
      );

      return {
        user: tokenData?.did || null,
      };
      return {
        user: null,
      };
```

jrhender (commented):
Looks really good, just a couple comments

src/modules/auth/login.controller.ts
Harasz: `verifyAccessToken` will throw an error when token is invalid, expired etc.

jrhender (changes requested):
@Harasz Thanks for fixing the typo. I think the commit message you used though of `Update test/auth/authStatus.testSuite.ts` would fail `commitlint` check. (Though I need to re-enable it 😅  https://github.com/energywebfoundation/iam-cache-server/pull/263)
Could you update the commit message please?

src/modules/auth/login.controller.ts
jrhender: Ok, thanks for the explanation. IMO, could be good to have a comment to explain that but it's fine for now 😄

Harasz:
Don’t worry it will be done

jrhender (approved):
Looks good to me!

PR #259, opened 2021-12-28
fix(organization): fix returning sub-orgs by a given parent namespace

Harasz:


src/modules/organization/organization.service.spec.ts
```diff
@@ -183,4 +184,25 @@ describe('OrganizationService', () => {
...
+      expect(subOrgs.length).toBe(organizations.length);
+      expect(subOrgs).toEqual(
+        organizations.map(org => {
+          delete org.parentOrg;
+          return org;
+        }),
+      );
+    });
+
+    it('getSubOrgs() it should throw an error for unknown pare organization', async () => {
```
JGiter, line 201:
```suggestion
    it('getSubOrgs() it should throw an error for unknown parent organization', async () => {
```

src/modules/organization/organization.service.ts
```diff
@@ -62,6 +62,8 @@ export class OrganizationService {
         where: {
           namespace,
         },
+        select: ['id', `subOrgs`],
```
JGiter, line 65: Is `subOrgs` needed here? I though that `select` should include only properties

JGiter (approved):

PR #260, opened 2021-12-28
feat(DID): refresh DID document from chain

Harasz:


src/modules/did/did.service.ts
```diff
@@ -151,19 +151,10 @@ export class DIDService {
...
-        cachedDIDDocument.id,
-        this.parseLogs(cachedDIDDocument.logs),
-      );
-
-      const updatedDidDocument = this.resolveDocumentFromLogs(
-        cachedDIDDocument.id,
-        logs,
-      );
-
+      const logs = await this.getAllLogs(did);
```
JGiter, line 154: Full sync by the time potentially might take enormous amount of time and number of requests to chain. Perhaps it should be one-time operation possibly triggered from some endpoint

Harasz: I see two ways:
1. Special script for that
2. Endpoint by only authorized for user with some specific role claim

JGiter: Also may be it is possible to add `readBatch` in did-reg

jrhender: I agree with Dmitry that we shouldn't change the approach permanently to be a full sync instead of incremental. IMO, a good approach would be if we can control via configuration whether or not a full vs incremental sync of DID docs is performed.

jrhender (changes requested):
We shouldn't completely remove the incremental sync approach as a part of this story

Harasz:
@JGiter @jrhender What do you think guys about current flow?

JGiter:
> @JGiter @jrhender What do you think guys about current flow?


Does this mean that production will always sync full?

Harasz:
No, we can specify if we want to turn on incremental or full sync for DID document via ENV variables.

src/modules/did/did.service.ts
```diff
@@ -142,11 +142,11 @@ export class DIDService {
   }
 
   /**
-   * Refresh the DID Document cache for a given DID.
+   * Fetch updates from IPFS for a given DID document.
```
jrhender, line 145: The DID document updates are obtained from the blockchain, not IPFS
```suggestion
   * Add any incremental changes to the DID document that occurred since the last sync.
```

src/modules/did/did.service.ts
```diff
@@ -179,6 +179,37 @@ export class DIDService {
...
+        ...updatedDidDocument,
+        service: updatedServices,
+        logs: JSON.stringify(logs),
+      });
+
+      return this.didRepository.save(updatedEntity);
+    } catch (err) {
+      this.logger.error(err);
+    }
+  }
```
jrhender, line 211: I don't think it is necessary to retrieve the previously cached document (i.e. this line `const cachedDIDDocument = await this.didRepository.findOne(did);`) as the `id` of the entity is the `did` in question so if an entity exists, I think we can just replace it. If this is the case, then this new `fullRefreshCachedDocument` is essentially the same as `addCachedDocument`, so I think the two methods should be combined somehow.

jrhender (changes requested):
IMO, it's moving in the right direction but I think the duplication between `fullRefreshCachedDocument` and `addCachedDocument` should be resolved.

JGiter (approved):

jrhender (approved):
Looks great, thanks!

PR #264, opened 2022-01-06
fix(claim): only verify issuer signature for issueToken

nichonien:
Only issuer signature for issuedToken needs to be verified. Remove wrong implementation for onChainProof verification.


jrhender (approved):
Thanks Ashish!

PR #281, opened 2022-01-13
Task/icl 128 add shared configs

Harasz:


JGiter (approved):

package.json
```diff
@@ -48,7 +48,10 @@
     "@energyweb/iam-contracts": "^3.6.0",
     "@ensdomains/ens": "^0.4.5",
     "@ensdomains/resolver": "^0.2.4",
+    "@ethersproject/abstract-provider": "5.4.1",
+    "@ethersproject/abstract-signer": "5.4.1",
```
JGiter, line 52: Why is this?

Harasz: because during `npm install` this sub-dependencies broke and did not compatible with existing `ethers` package. So I have to add this dependencies manually to set the proper version.

JGiter (approved):

PR #282, opened 2022-01-14
bug/can't get asset by

JGiter:


JGiter:
Fixed in https://github.com/energywebfoundation/iam-cache-server/pull/289

PR #291, opened 2022-01-18
ci: add argocd project

hejkerooo:


jrhender (changes requested):
Thanks for the PR. Could you explain a bit how this relates to https://github.com/energywebfoundation/iam-cache-server-helm/ 
I thought the helm stuff was separate from the app code?

hejkerooo:
>Thanks for the PR. Could you explain a bit how this relates to https://github.com/energywebfoundation/iam-cache-server-helm/
I thought the helm stuff was separate from the app code?

@jrhender - our goal is to separate helm repository from values repository as we want to make Helm repository public in the future and make it easier for developers to change app. values

jrhender (approved):
Looks good to me!

PR #294, opened 2022-01-18
Sync development with master

JGiter:
This needs to unblock https://github.com/energywebfoundation/iam-cache-server/pull/292


src/modules/claim/claim.controller.ts
```diff
@@ -238,7 +238,7 @@ export class ClaimController {
       'should return claims for namespaces like ' +
       '`admin.roles.myApp.apps.myOrg.iam.ewc`',
   })
-  public async getByParentNamespace(@Param('id') id: string) {
+  public async getByParentNamespace(@Param('namespace') id: string) {
```
jrhender, line 241: Is `namespace` the correct value?

jrhender (approved):
Approved with one question

src/modules/claim/claim.controller.ts
JGiter: If I am not mistaken it is local identifier, which is not part of API. `namespace` is also field of the `RoleClaim`, so it is looks consistent

JGiter: Although there is still some bug with this endpoint, as it is always returns [ ] though locally everything works fine for `sync_dev_with_master` branch

PR #304, opened 2022-01-26
Update staging cache-server

jrhender:
I propose we promote `dev` cache-server to `staging`. There are no breaking changes, that I am aware of.


Harasz (approved):
Are you sure that those variables are set on staging? `DB_MAXIMUM_CONNECTION_POOL`, ` 'DB_CONNECTION_TIMEOUT`

jrhender:
> Are you sure that those variables are set on staging? `DB_MAXIMUM_CONNECTION_POOL`, ` 'DB_CONNECTION_TIMEOUT`

Good question, they might not be. I don't see them in the `staging` file values file, so I guess they aren't. I think we can merge anyways as there are default values set.

@hejkerooo How would we go about setting them?

JGiter (approved):

PR #305, opened 2022-01-26
Task/ics 120 add support on chain without request

Harasz:
Add a support for on-chain role types without a previous request. Also, refactor claim service (split into smaller parts).


package.json
```diff
@@ -94,12 +94,13 @@
     "graphql-tools": "^7.0.4",
     "grpc": "^1.24.6",
     "helmet": "^4.4.1",
+    "iam-client-lib": "^4.0.1",
```
jrhender, line 97: I don't think we should add ICL as this will create a circular dependency

jrhender (changes requested):

src/modules/claim/services/claim-issuance.service.ts
```diff
@@ -0,0 +1,150 @@
...
+    return ClaimHandleResult.Failure('claim issuance criteria not met');
+  }
+
+  private async handleClaimWithoutPreviousRequest(
+    rq: IClaimIssuance
+  ): Promise<ClaimHandleResult> {
+    let claimType: string;
+    let claimTypeVersion: string;
+
+    if (rq.issuedToken) {
```
JGiter, line 52: Probably this can be omitted as claim issued without request should include `claimType` and `claimTypeVersion` https://github.com/energywebfoundation/iam-client-lib/blob/2ff0fa4c4a2d44cf13f8a919352932791154ebac/src/modules/claims/claims.service.ts?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L361

src/modules/claim/services/claim-issuance.service.ts
```diff
@@ -0,0 +1,150 @@
...
+    await this.createAndIssue(dto, dto.requester);
+
+    return ClaimHandleResult.Success();
+  }
+
+  private async handleClaimWithPreviousRequest(
+    rq: IClaimIssuance,
+    previouslyRequestedClaim: RoleClaim
+  ): Promise<ClaimHandleResult> {
+    await this.roleService.verifyEnrolmentIssuer({
```
JGiter, line 98: Exact verification is also in `handleClaimWithoutPreviousRequest`. Let's hoist it

src/modules/claim/services/claim-issuance.service.ts
```diff
@@ -0,0 +1,150 @@
...
+  /**
+   * Handles claim issuance request saving and updates.
+   * Two scenarios are handled - issue requested claim and issue not-requested claim
+   * @param rq IClaimIssuance request
+   */
+  public async handleClaimIssuanceRequest(
+    rq: IClaimIssuance,
+    previouslyRequestedClaim: RoleClaim | undefined
+  ): Promise<ClaimHandleResult> {
+    if (!previouslyRequestedClaim && (rq.issuedToken || rq.onChainProof)) {
```
JGiter, line 33: `rq.issuedToken || rq.onChainProof` is checked in multiple places. I believe this can be checked before issuance

src/modules/claim/services/claim-issuance.service.ts
```diff
@@ -0,0 +1,150 @@
...
+  ) {
+    this.logger.setContext(ClaimIssuanceService.name);
+  }
+
+  /**
+   * Handles claim issuance request saving and updates.
+   * Two scenarios are handled - issue requested claim and issue not-requested claim
+   * @param rq IClaimIssuance request
+   */
+  public async handleClaimIssuanceRequest(
```
JGiter, line 29: IMO there is not much need in this method. We can call one of the handles right after getting `previouslyRequestedClaim`

JGiter (changes requested):
Thanks @Harasz . Looks good, but there is some room to improve code quality

.github/workflows/cache_server.test.yml
```diff
@@ -56,9 +56,6 @@ jobs:
       - name: Generate keys
         run: npm run generate:jwtkeys
 
-      - name: Check linting
-        run: npm run lint
```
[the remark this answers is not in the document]
Harasz: We have a separate workflow for it.
https://github.com/energywebfoundation/iam-cache-server/blob/develop/.github/workflows/codelint.yml

src/modules/claim/services/claim-issuance.service.ts
Harasz: Currently is just for On-chain claims
https://github.com/energywebfoundation/iam-client-lib/blob/2ff0fa4c4a2d44cf13f8a919352932791154ebac/src/modules/claims/claims.service.ts?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L391

src/modules/claim/services/claim-issuance.service.ts
Harasz: You mean something like that `const isValidRequest = rq.issuedToken || rq.onChainProof`

src/modules/claim/services/claim-issuance.service.ts
Harasz: It will complicated the code from my perspective. Maybe you can suggest the change here

src/modules/claim/services/claim-issuance.service.ts
JGiter: Ah, right. I guess will need to add it to off-chain claim as well

Harasz: It will clarify logic a little bit for sure

src/modules/claim/services/claim-issuance.service.ts
JGiter: I would do like this:
in `handleClaimIssuanceRequest`
- make sure that `issuedToken` or `onChainProof` is present
- make sure that `claimType` and `claimTypeVersion` are present, otherwise extract them from `issuedToken` or throw
- `verifyEnrollmentIssuer`, `verifyEnrollmentPrecodnditions`
- depending on existence of `previousyRequestedClaim` call `issue` or `createAndIssue`

JGiter (approved):

src/modules/claim/services/claim-issuance.service.ts
Harasz: Okay, lets see how it will look like

src/modules/claim/claim.controller.ts
```diff
@@ -121,10 +125,11 @@ export class ClaimController {
     );
 
     this.logger.debug(`credentials issued by ${did}`);
+    return result;
   }
 
   @Post('/request/:did?')
-  // @ApiExcludeEndpoint()
+  @ApiExcludeEndpoint()
```
jrhender, line 132: @Harasz Do you know why we had this commented before? It seems strange to me that it is excluded from the docs

src/modules/claim/claim.dto.ts
```diff
@@ -124,10 +151,8 @@ export class NewClaimIssueDTO extends ClaimIssueDTO {
   claimTypeVersion: string;
 
   @IsEnum(RegistrationTypes, { each: true })
-  // Optional so as to not break existing clients. Can be made mandatory in future.
-  @IsOptional()
   @ApiProperty()
-  registrationTypes: [RegistrationTypes];
+  registrationTypes: [RegistrationTypes, RegistrationTypes?];
```
jrhender, line 155: Isn't `RegistrationTypes?` superset of `RegistrationTypes`. I.e. it is `RegistrationTypes` + `undefined\null`?
```suggestion
  registrationTypes: [RegistrationTypes?];
```

src/modules/claim/claim.types.ts
```diff
@@ -25,6 +25,8 @@ export interface IClaimRequest extends IMessage {
 export interface IClaimIssuance extends IMessage {
   issuedToken?: string;
   onChainProof?: string;
+  claimType?: string;
+  claimTypeVersion?: string;
```
jrhender, line 29: Why are these props optional? Doesn't every issued claim have a `claimType` + `claimTypeVersion` at this point?

jrhender (commented):
Just a few small comments, but looks pretty good

src/modules/claim/claim.controller.ts
Harasz: Yeah, last time I comment this line out and forgot to undo. So it's back to original state

src/modules/claim/claim.dto.ts
Harasz: It;s ideally present how the object should look like, at lest one element is required but array can contain max two elements. Previously we can have a object with a five registration types, which is not allowed, currently TS will catch this situation.

src/modules/claim/claim.types.ts
Harasz: Unfortunately no, just on-chain claim without previous request has these props.
https://github.com/energywebfoundation/iam-client-lib/blob/2ff0fa4c4a2d44cf13f8a919352932791154ebac/src/modules/claims/claims.service.ts#L392

jrhender (approved):
Looks good, thanks!

PR #310, opened 2022-01-28
ci: set additional tag for image

hejkerooo:
This PR will fix deployments for OCI based helm charts

### DIFF
--- a/.github/workflows/build_and_deploy.yml
+++ b/.github/workflows/build_and_deploy.yml
@@ -87,3 +87,9 @@ jobs:
         with:
           command: app set iam-cache-server
           options: -p image.tag=${{needs.unique_id.outputs.unique_id}}
+      - name: ArgoCD overvrite values.yaml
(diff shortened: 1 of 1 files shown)


Harasz (approved):

.github/workflows/build_and_deploy.yml
```diff
@@ -87,3 +87,9 @@ jobs:
         with:
           command: app set iam-cache-server
           options: -p image.tag=${{needs.unique_id.outputs.unique_id}}
+      - name: ArgoCD overvrite values.yaml
```
jrhender, line 90: @hejkerooo Could you add a bit to this suggested comment to explain the why?
```suggestion 
      # For OCI based helm charts, tag needs to be overwritten.. {why??}
      - name: ArgoCD overwrite values.yaml
```

jrhender (commented):
@hejkerooo I don't really understand it why this change is necessary. Could you maybe link to some documentation or explain a bit more?

PR #313, opened 2022-02-01
docs: add list with breaking changes

Harasz:


Harasz:
image
View of the compiled file.

JGiter:
I suggest to call this `Migration Guide` and put in the docs. This is the way how it is done in `ethers`

Harasz:
It should be added docs folder, but one thing concerns me: `docs` dir is auto generated, so It would be better to add it to root dir and then copy during generation to `docs` dir. What you think? @JGiter @jrhender

jrhender:
@Harasz I thought that only the `api` dir in the `docs` dir is autogenerated. I think the rest is static. Is this not right?

Harasz:
Ohh yeah, you're right. I will move to docs and change to migration guide. Thanks  guys!

jrhender (approved):

PR #314, opened 2022-02-02
feat: improve sentry tracing

Harasz:


src/modules/did/did.service.ts
```diff
@@ -124,23 +124,41 @@ export class DIDService implements OnModuleInit {
...
    */
-  public async addCachedDocument(did: string) {
-    const transaction = this.sentryTracingService.startTransaction(
-      'process-did',
-      'Process DID',
-      {
-        did,
-      }
-    );
+  public async addCachedDocument(did: string, isSync = false) {
```
JGiter, line 127: I am a bit confused here. I thought that `addCachedDocument` should be used only to add document in cache. Am I wrong?

JGiter (approved):
Looks good. Only a little notice about additional argument to `addCacheDocument`

src/modules/did/did.service.ts
Harasz: It was, right now it is also used in full mode sync since the steps are identically for this cases

PR #341, opened 2022-02-28
task/ICL-251 update domain resolver to version 2

JGiter:
https://energyweb.atlassian.net/browse/ICL-251

merge after https://github.com/energywebfoundation/iam-client-lib/pull/478


.env.ci
```diff
@@ -20,12 +20,12 @@ CHAIN_NAME=volta
 
 # ENS
 ENS_URL=https://volta-rpc-vkn5r5zx4ke71f9hcu0c.energyweb.org/
-PUBLIC_RESOLVER_ADDRESS=0x0a97e07c4Df22e2e31872F20C5BE191D5EFc4680
```
jrhender, line 23: I don't think we want to remove this yet, also in `.env.dist`

jrhender (changes requested):

jrhender (changes requested):
@JGiter I think the actual location that matters now is https://github.com/energywebfoundation/ssi-hub/blob/develop/devops/ so these addresses should be added there as well

jrhender (approved):
Thanks @JGiter

jrhender:
> Got it. So `.env.*` are not used anymore?

Hmm, I'm not sure. Maybe we should just keep a single `.env` there for local development

jrhender:
> > Got it. So `.env.*` are not used anymore?
> 
> Hmm, I'm not sure. Maybe we should just keep a single `.env` there for local development

I made task here https://energyweb.atlassian.net/browse/ICS-189

PR #348, opened 2022-03-09
bug/ICS-192 incorrect domain controller

JGiter:
https://energyweb.atlassian.net/browse/ICS-192

### DIFF
--- a/src/modules/ens/ens.service.ts
+++ b/src/modules/ens/ens.service.ts
@@ -123,7 +123,6 @@ export class EnsService implements OnModuleDestroy {
       this.InitEventListeners();
       this.syncENS();
     }
-    this.syncENS();
   }
 
   private async deleteNamespace(hash: string) {
--- a/src/modules/ens/ens.service.ts
+++ b/src/modules/ens/ens.service.ts
@@ -27,6 +27,7 @@ import { Logger } from '../logger/logger.service';
 import { Provider } from '../../common/provider';
 import { SentryTracingService } from '../sentry/sentry-tracing.service';
 
+const { solidityKeccak256 } = utils;
 export const emptyAddress = '0x'.padEnd(42, '0');
 
 @Injectable()
@@ -122,6 +123,7 @@ export class EnsService implements OnModuleDestroy {
       this.InitEventListeners();
       this.syncENS();
     }
+    this.syncENS();
   }
 
   private async deleteNamespace(hash: string) {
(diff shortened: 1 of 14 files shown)


src/modules/ens/ens.service.ts
```diff
@@ -164,8 +166,11 @@ export class EnsService implements OnModuleDestroy {
     });
 
     // Register event handler for owner change or namespace deletion
-    this.ensRegistry.on('NewOwner', async (node, _, owner) => {
-      await this.eventHandler({ hash: node, owner });
+    this.ensRegistry.on('NewOwner', async (node, label, owner) => {
+      await this.eventHandler({
+        hash: solidityKeccak256(['bytes', 'bytes'], [label, node]),
```
jrhender, line 171: I don't really get this 😓 . So there was a bug here before? You need to hash both the `label` and the `node` together? I thought that `node` was the namehash of the domain

JGiter: `eventHandler` is expected to be given hash of the transferred domain, which consist from label and node. I think we haven't discovered this bug yet because `NewOwner` event is emitted by `setSubnodeOwner` which we haven't used yet, although this method is preset in iam-lib

jrhender: Hmm, interesting, thanks. Do you think it would make sense to use https://docs.ethers.io/v5/api/utils/hashing/#utils-namehash for this rather than `solidityKeccak256` directly? (sorry if it's a dumb suggestion 😆 )

src/modules/ens/ens.service.ts
```diff
@@ -122,6 +123,7 @@ export class EnsService implements OnModuleDestroy {
       this.InitEventListeners();
       this.syncENS();
     }
+    this.syncENS();
```
Harasz, line 126: Don't forget to remove it

JGiter: thanks, will remove

src/modules/ens/ens.service.ts
jrhender: Right, that makes sense 👍  So will we need to trigger this event again for the previously migrated domains? Or will the daily ENS sync fix it you think?

JGiter: I think daily sync should fix this

jrhender (approved):

PR #352, opened 2022-03-11
bug/ICS-192 incorrect subdomain hash

JGiter:


src/modules/ens/ens.service.ts
```diff
@@ -167,7 +167,7 @@ export class EnsService implements OnModuleDestroy {
     // Register event handler for owner change or namespace deletion
     this.ensRegistry.on('NewOwner', async (node, label, owner) => {
       await this.eventHandler({
-        hash: solidityKeccak256(['bytes', 'bytes'], [label, node]),
+        hash: solidityKeccak256(['bytes32', 'bytes32'], [node, label]),
```
jrhender, line 170: Can't hurt to have a link to the algorithm, IMO.
```suggestion
        // https://docs.ens.domains/contract-api-reference/name-processing#algorithm
        hash: solidityKeccak256(['bytes32', 'bytes32'], [node, label]),
```

jrhender (approved):
Look like this is now the correct order 👍

PR #353, opened 2022-03-11
feat: store request claim origin

Harasz:


src/modules/claim/entities/roleClaim.entity.ts
```diff
@@ -12,6 +12,14 @@ export class RoleClaim implements IRoleClaim {
   static create(data: Partial<RoleClaim>): RoleClaim {
     const entity = new RoleClaim();
     const jwt = new JWT(new Keys());
+
+    if (
+      data?.requestCompletionRedirectUrl &&
+      !isURL(data.requestCompletionRedirectUrl, { require_tld: false })
```
jrhender, line 18: `require_tld` is for `localhost`, right? Maybe add comment about this?
```suggestion
      // https://github.com/validatorjs/validator.js/issues/754#issuecomment-346826183
      !isURL(data.requestCompletionRedirectUrl, { require_tld: false })
```

src/modules/claim/services/claim.service.ts
```diff
@@ -62,7 +62,8 @@ export class ClaimService {
    * @param rq IClaimRequest request
    */
   public async handleClaimEnrolmentRequest(
-    rq: IClaimRequest
+    rq: IClaimRequest,
+    url: string
```
jrhender, line 66: I would consider carrying the `redirectUri` name through here
```suggestion
    redirectUri: string
```

src/modules/claim/entities/roleClaim.entity.ts
```diff
@@ -83,6 +91,10 @@ export class RoleClaim implements IRoleClaim {
   @Field()
   @Column()
   namespace: string;
+
+  @Field({ nullable: true })
+  @Column({ nullable: true })
+  requestCompletionRedirectUrl?: string;
```
jrhender, line 97:
```suggestion
  /**
   * This is location to which the client can be redirected after the request/exchange
   * TODO: Confirm the extent to which this is similar to OpenID Connect redirect_uri
   */
  @Field({ nullable: true })
  @Column({ nullable: true })
  redirectUri?: string;
```

src/modules/claim/services/claim.service.ts
```diff
@@ -111,14 +112,16 @@ export class ClaimService {
    */
   public async create(
     data: ClaimRequestDTO,
-    subject: string
+    subject: string,
+    url: string
```
jrhender, line 116:
```suggestion
    redirectUri: string
```

jrhender (commented):
Looks good. Just some minor suggestions

src/modules/claim/entities/roleClaim.entity.ts
Harasz: I am thinking right now that is not exactly the `redirectionUri` but `requestOriginUri`. One of the potential use cases is redirection, but not necessary the only one. What you think?

JGiter: Although both points are valid I am inclined to `redirectUri`, because the purpose of this property is to redirect. It is just happened that this value is set to origin

JGiter (approved):

src/modules/claim/entities/roleClaim.entity.ts
jrhender: I agree with Dmitry, I think that the property is currently coming from HTTP Origin header, but we could set it in another way in the future if necessary

jrhender (approved):
Looks good!

PR #368, opened 2022-03-23
feat: add `decentralized-web-node` module

Harasz:


JGiter (changes requested):

docs/api/classes/ethers_StakingPoolFactory.StakingPoolFactory.md
```diff
@@ -0,0 +1,602 @@
+# Class: StakingPoolFactory
```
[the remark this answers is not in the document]
Harasz: I will clean docs before merge

src/common/array-distinct.decorator.ts
```diff
@@ -0,0 +1,31 @@
+// https://github.com/typestack/class-validator/issues/592#issuecomment-621645012
+
+import {
+  registerDecorator,
+  ValidationArguments,
+  ValidationOptions,
+} from 'class-validator';
+
+export default function ArrayDistinct(validationOptions?: ValidationOptions) {
```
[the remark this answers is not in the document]
Harasz: Hah, I was looking for such decorator, thanks for letting my know

src/modules/claim/services/claim-issuance.service.ts
```diff
@@ -133,6 +133,7 @@ export class ClaimIssuanceService {
       subject,
       namespace: parent,
       isAccepted: true,
+      vp: data?.vp ? JSON.parse(data.vp) : undefined,
```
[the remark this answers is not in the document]
Harasz: It is, holder will send it to SSI-Hub as a JSON string (for simplicity) and then I need to resolve it to actual object before saving to database

src/modules/verifiable-credential/services/vc-match.service.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    const vcStorage = await this.vcStorageService.findAll(holder);
+    const credentials = vcStorage.map((vc) => vc.vc);
+
+    const srMatches = pex.selectFrom(presentationDefinition, credentials, [
+      holder,
+    ]);
+
+    if (srMatches.errors.length) {
+      const errorMessage = `Error while matching presentation definition with VCs: ${this.checkedToString(
```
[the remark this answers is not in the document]
Harasz: I think that kind of message can be misleading and it don't tell the user what is the actual error (harder to debug)

src/modules/verifiable-credential/services/vc-match.service.ts
```diff
@@ -0,0 +1,63 @@
...
+    return checked
+      .filter((err) => err.message)
+      .map((err) => err.message)
+      .join(', ');
+  }
+
+  async matchCredentials(
+    presentationDefinition: PresentationDefinitionDto,
+    holder: string
+  ): Promise<VerifiableCredential<CredentialSubject>[]> {
```
jrhender, line 33: I think that instead of returning the `Credentials` we should just return the `SelectResults`. This is because the client will need more than just the credentials to display to the user. For instance, I think `SelectResult.matches` is required so that the UI can guide the client as to which credentials can meet which submission requirement

src/modules/verifiable-credential/verifiable-credential.controller.ts
```diff
@@ -0,0 +1,83 @@
...
+    await this.vcStorageService.save({
+      vcId: vc.id,
+      issuer: vc.issuer,
+      subject: vc.credentialSubject.id,
+      holder: user,
+      vc,
+    });
+    this.logger.log(`Saved verifiable credential ${vc.id}`);
+  }
+}
```
jrhender, line 83: Instead of this custom REST interface, I think we might as well try to use the DIF Decentralized Web Node interface (or another standard storage interface if you know one), even if it will take a little bit more time to implement. I will make a brief PR today to sketch out what this could look like

src/modules/claim/services/claim-issuance.service.ts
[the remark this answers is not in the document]
JGiter: I was confused that holder will call `issue` endpoint, but actually it is correct

src/modules/verifiable-credential-storage/entities/vc-storage.entity.ts
```diff
@@ -0,0 +1,33 @@
+import {
+  CredentialSubject,
+  VerifiableCredential,
+} from '@ew-did-registry/credentials-interface';
+import { Column, PrimaryGeneratedColumn, Entity } from 'typeorm';
+import { IVCStorage } from '../types/vc-storage';
+
+@Entity()
+export class VCStorage implements IVCStorage {
```
jrhender, line 9: IMO, this module shouldn't really know anything about VCs. I would therefore change this entity to be a `CollectionsObject` and just have properties that are generic to `CollectionsObjects` (`schema`, `dataFormat`, `dateCreated`, `datePublished`) https://identity.foundation/decentralized-web-node/spec/#write

In the case of VCs, I would then set the properties as:
- `schema`: An array containing the `types` of the VCs https://www.w3.org/TR/vc-data-model/#types
- `dataFormat`: `application/json` or `application/ld+json` (if it is a JSON-LD VC https://www.w3.org/TR/json-ld/#interpreting-json-as-json-ld)
- `dateCreated`: the UNIX Epoch timestamp that the object was initially stored
- `datePublished`: I think we just leave this blank for now?

jrhender (changes requested):
I would recommend a rename of the modules as:
- `verifiable-credential-storage` -> `decentralized-web-node` (or maybe `decentralized-web-node-collections`)
- `verifiable-presentation` -> `presentation-exchange` (to make it clear that is providing endpoint related to the `presentation-exchange` specification)
What do you think of this?

src/modules/decentralized-web-node/dtos/verifiable-credential.dto.ts
```diff
@@ -0,0 +1,79 @@
...
+} from 'class-validator';
+import {
+  CredentialSubject,
+  VerifiableCredential,
+} from '@ew-did-registry/credentials-interface';
+import { Type } from 'class-transformer';
+import { CredentialSchemaOrCredentialStatusDto } from './credential-schema-status.dto';
+import { ProofDto } from './proof.dto';
+
+export class VerifiableCredentialDto
```
jrhender, line 17: Can this class be removed? It doesn't look like it is referenced anywhere. Then `proof` and `eip712-domain` could be removed as well?

src/modules/decentralized-web-node/dtos/match-results.dto.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  @ApiProperty({ enum: ['info', 'warn', 'error'] })
+  status: Status;
+
+  @ApiProperty({ required: false })
+  message?: string;
+}
+
+export class SubmissionRequirementMatchDto
+  implements SubmissionRequirementMatch
```
jrhender, line 22: This can be removed from `decentralized-web-node`, right?

jrhender (commented):
Will look some more, a couple quick clean up comments to start

src/modules/decentralized-web-node/services/vc-storage.service.ts
```diff
@@ -0,0 +1,78 @@
...
+    });
+    if (foundedMessage) {
+      throw new ConflictException(
+        `Message with id ${message.objectId} already exists`
+      );
+    }
+
+    const item = await this.messageRepository.save(message);
+    return item;
+  }
```
jrhender, line 77: I think this logic is incorrect. I think it is possible in principle to update an object. I think rather the `dateCreated` of the previous message should be checked and the object should be updated is the `dateCreated` of the message is newer. https://identity.foundation/decentralized-web-node/spec/#processing-instructions

jrhender (changes requested):
Still reviewing but adding a couple other comments

src/modules/decentralized-web-node/decentralized-web-node-response.interceptor.ts
```diff
@@ -0,0 +1,38 @@
...
+    return next.handle().pipe<ResponseObjectDto>(
+      catchError(async (err) => {
+        const status = err.status || 500;
+        console.log(err);
+        response.status(status);
+        return {
+          requestId: request.body.requestId || '',
+          status: {
+            code: status,
+            message: err.message,
```
jrhender, line 31: Spec says that the message should be "The request could not be processed correctly". What do you think of this? To me it makes sense, as the server shouldn't necessarily reveal the internal error to the client.
https://identity.foundation/decentralized-web-node/spec/#general-request-processing-error
```suggestion
            message: "The request could not be processed correctly",
```

jrhender (commented):
Some comments about the interceptor. Still reviewing 😄

src/modules/decentralized-web-node/decentralized-web-node-response.interceptor.ts
```diff
@@ -0,0 +1,38 @@
...
+    context: ExecutionContext,
+    next: CallHandler
+  ): Observable<ResponseObjectDto> {
+    const request: Request = context.switchToHttp().getRequest();
+    const response: Response = context.switchToHttp().getResponse();
+
+    return next.handle().pipe<ResponseObjectDto>(
+      catchError(async (err) => {
+        const status = err.status || 500;
+        console.log(err);
```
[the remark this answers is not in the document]
Harasz: It's just for test here. Error logging should be handled by `SentryErrorInterceptor`

src/modules/decentralized-web-node/entities/message.entity.ts
```diff
@@ -0,0 +1,52 @@
...
+    const entity = new Message();
+    Object.assign(entity, data);
+    return entity;
+  }
+
+  @PrimaryColumn()
+  objectId: string;
+
+  @Column()
+  createdBy: string;
```
jrhender, line 18: Is `createdBy` a part of the spec? I think it would be better to use `target` here?
The spec doesn't really specify `target` in text, but it's in all of the examples: https://identity.foundation/decentralized-web-node/spec/#messages
```suggestion
  target: string;
```

src/modules/decentralized-web-node/services/message.service.ts
```diff
@@ -0,0 +1,115 @@
...
+        objectId: message.objectId,
+      },
+    });
+    let item: Message = foundedMessage;
+
+    if (foundedMessage && foundedMessage.createdBy !== user) {
+      throw new UnauthorizedException(
+        `User is not allowed to update message ${message.objectId}`
+      );
+    }
```
jrhender, line 87: Does still code mean that the implementation can't store a message with the same `objectId` but created by different users?
I think different users could store the same object, we should instead authorize at the `target` level

jrhender (commented):
Some further comments but it's looking good!

src/modules/decentralized-web-node/decentralized-web-node.controller.ts
```diff
@@ -0,0 +1,43 @@
...
+    private readonly collectionHandlerService: CollectionHandlerService
+  ) {
+    this.logger.setContext(DecentralizedWebNodeController.name);
+  }
+
+  @ApiOkResponse({
+    type: ResponseObjectDto,
+  })
+  @Post('/')
+  public async credentialsStorageEntry(
```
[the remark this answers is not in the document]
Harasz: I'd call it `requestHandler`

src/modules/decentralized-web-node/entities/message.entity.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  @Column({ nullable: true })
+  data?: string;
+
+  @Column('jsonb')
+  message: MessageDto;
+
+  @Column('jsonb')
+  descriptor: CollectionsQueryDescriptorDto | CollectionsWriteDescriptorDto;
+
```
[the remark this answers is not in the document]
Harasz: `descriptor` shouldn't be stored, I made that by mistake. For message I used this interface bc we are saving exactly what comes from user to database.

src/modules/decentralized-web-node/entities/message.entity.ts
```diff
@@ -0,0 +1,52 @@
+import { Column, Entity, PrimaryColumn } from 'typeorm';
+import { CollectionsQueryDescriptorDto } from '../dtos/collections-query-descriptor.dto';
+import { CollectionsWriteDescriptorDto } from '../dtos/collections-write-descriptor.dto';
+import { MessageDto } from '../dtos/message.dto';
+
+@Entity()
+export class Message {
```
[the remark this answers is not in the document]
Harasz: Not exactly. I agree it's a better name, but we can avoid nullable of those properties, bc: `The object MAY contain a schema property,` (https://identity.foundation/decentralized-web-node/spec/#write) and `If the [Message](https://identity.foundation/decentralized-web-node/spec/#messages) has data associated with it, passed directly via the data property of the [Message](https://identity.foundation/decentralized-web-node/spec/#messages) object or through a channel external to the message object, the descriptor object MUST contain a dataFormat property,` (https://identity.foundation/decentralized-web-node/spec/#message-descriptors)

src/modules/decentralized-web-node/services/message.service.ts
Harasz: ObjectId is the uuid v4 version, do you think it should be the same for all objects?

jrhender: Hmm, sorry, maybe I confused. But what I'm trying to say is that we should allow multiple `targets` to store the same object and I don't think this code allows this because it assumes that there is only one of the message with a given `objectId` (because it uses `findOne` and doesn't filter on `target`). Is this not how you see it?

src/modules/decentralized-web-node/entities/message.entity.ts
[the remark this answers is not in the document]
jrhender: Ah okay, I think you meant that we "can't" avoid nullable, right? Maybe you're right, and we need to keep them nullable. Maybe we just change the name then?

src/modules/decentralized-web-node/entities/message.entity.ts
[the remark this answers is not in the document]
jrhender: Ah okay, you're thinking that is best to have a kind of "backup" of the message in the `message` property? Or I guess it is handy to have because it is what we return in response to a `Query`?

src/modules/decentralized-web-node/decentralized-web-node.controller.ts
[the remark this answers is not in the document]
jrhender: You're right, that's better

src/modules/decentralized-web-node/services/message.service.ts
Harasz: Anyone can create own `objectId` it can be different for the same message data

src/modules/decentralized-web-node/entities/message.entity.ts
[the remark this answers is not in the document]
Harasz: The second option: `it is handy to have because it is what we return in response to a Query`

src/modules/decentralized-web-node/entities/message.entity.ts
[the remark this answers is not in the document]
Harasz: Yeah soryy, I meant `can't`. The name will be changed

src/modules/decentralized-web-node/services/message.service.ts
jrhender: Right, I'm imagining a kind of multicast messaging scenario where multiple targets (all using the same decentralized web node server) receive the same message (so it would have the same `objectId`). I think this code would fail in this case, right?

Harasz: Ohh, okay I see. let me fix that

src/modules/decentralized-web-node/services/handler.service.ts
```diff
@@ -0,0 +1,143 @@
...
+import { MessageDto } from '../dtos/message.dto';
+import { ResponseReplyDto } from '../dtos/response-reply.dto';
+import { CollectionsWriteMessage } from '../entities/collections-write-message.entity';
+import { MessageService } from './message.service';
+
+@Injectable()
+export class CollectionHandlerService {
+  constructor(
+    private readonly logger: Logger,
+    private readonly messageService: MessageService
```
jrhender, line 19: Maybe it's just me, but I find the naming/relationship between this class and `MessageService` to be pretty confusing. I guess we should start by changing the name of this class to align with the filename.

In your view, why are the `HandlerService` and `MessageService` separate? It is nice to have smaller classes but the distinction between them isn't really clear to me

```suggestion
export class HandlerService {
  constructor(
    private readonly logger: Logger,
    private readonly messageService: MessageService
```

src/modules/verifiable-presentation/services/vc-match.service.ts
```diff
@@ -0,0 +1,32 @@
...
+    presentationDefinition;
+    holder;
+    const pex: PEX = new PEX();
+    pex;
+    this.messageService;
+    // const vcStorage = await this.vcStorageService.findAll(holder);
+    // const credentials = vcStorage.map((vc) => vc.vc);
+
+    // return pex.selectFrom(presentationDefinition, credentials, [holder]);
+    return [] as any;
```
jrhender, line 30: I guess this still needs to be wired up to the decentralized-web-node module. But the interface looks good!

jrhender (commented):
A couple more comments but I think everything is on the right track. Thanks @Harasz , great work!

src/modules/decentralized-web-node/services/handler.service.ts
Harasz: So, `MessageService` has a method that allow you to modify and work with `CollectionsWriteMessage` entities. It's more like repository I'd say. On the other hand, we have a service called `HandlerService` that is responsible just for receiving query/write messages and manage them.

src/modules/verifiable-presentation/services/vc-match.service.ts
Harasz: First I finish `decentralized-web-node` module then I will switch to this module. I was thinking about writing query messages and internally send it to `HandlerService`, what's your opinion on this approach?

jrhender: >  I was thinking about writing query messages and internally send it to HandlerService, what's your opinion on this approach?

This sounds good the me 👍

src/modules/decentralized-web-node/services/handler.service.ts
jrhender: Ok, thanks for the explanation. We can leave as is, I think it's better to have them separate than together 👍

Harasz: It's way readable and we can only share a `HandlerService` between modules, so no direct operation on entities, only via query/write message.

JGiter (approved):
Nice job @Harasz !

src/modules/decentralized-web-node/services/message.service.ts
```diff
@@ -0,0 +1,113 @@
...
+      order,
+    });
+    return items;
+  }
+
+  async save(
+    message: CollectionsWriteMessage,
+    user: string
+  ): Promise<CollectionsWriteMessage | undefined> {
+    const foundedMessage = await this.messageRepository.findOne({
```
jrhender, line 79:
```suggestion
    const foundMessage = await this.messageRepository.findOne({
```

src/modules/decentralized-web-node/services/message.service.ts
```diff
@@ -0,0 +1,113 @@
...
+      case foundedMessage && message.dateCreated === foundedMessage.dateCreated:
+        if (message.cid > foundedMessage.cid) {
+          item = await this.messageRepository.save(message);
+        } else {
+          item = foundedMessage;
+        }
+        break;
+
+      default:
+      case !foundedMessage:
```
jrhender, line 106: To me, it's weird to have a case statement after `default`. I think the `case` could be removed without changing the behaviour here, right?
```suggestion
      default:
```

src/modules/decentralized-web-node/services/message.service.ts
```diff
@@ -0,0 +1,113 @@
...
+        item = await this.messageRepository.save(message);
+        break;
+    }
+
```
jrhender, line 110: IMO, an `if` statement would be a better fit here. I personally find `case` statements a bit harder to read and I think a `case` statement doesn't make sense in this case as evidenced by the `switch(true)` (which is essentially meaningless)

.github/workflows/cache_server.test.yml
```diff
@@ -56,8 +56,9 @@ jobs:
       - name: Generate keys
         run: npm run generate:jwtkeys
 
-      - name: Audit package dependencies
-        run: npm audit --audit-level=critical
+      # TODO!: Remove before merge
+      # - name: Audit package dependencies
+      #   run: npm audit --audit-level=critical
```
jrhender, line 61: Maybe just check if this is still necessary before merging

jrhender (approved):
Looks good to me, great work

PR #371, opened 2022-04-07
feat: obtain my VCs which match a presentation definition 

Harasz:


jrhender (commented):
What do you think of renaming this module to `presentation-exchange` to be a bit more specific, as it seems that everything in the module has to do with the Presentation Exchange spec as the moment?

JGiter (approved):

src/modules/presentation-exchange/services/vc-match.service.ts
```diff
@@ -0,0 +1,75 @@
...
+
+    const entries = reply.entries
+      .map(this.decodeData)
+      .filter(Boolean)
+      .filter(isVerifiableCredential);
+
+    return entries;
+  }
+
+  async matchCredentials(
```
jrhender, line 66: Minor style point, but IMO, it's better to put public members of a class at the top as a rule of thumb. This is because it helps readers find the interface of the class quickly without having to look at implementation details.
Some discussion about this here. There definitely isn't clear agreement on this, but just my 2 cents https://stackoverflow.com/questions/1760850/best-practice-ordering-of-public-protected-private-within-the-class-definition

jrhender (approved):
A couple comments, but looks good, thanks!

PR #372, opened 2022-04-12
fix(claim): fix get by subject endpoint

Harasz:


src/modules/claim/services/claim.service.ts
```diff
@@ -172,7 +172,7 @@ export class ClaimService {
     }
 
     if (namespace) {
-      qb.andWhere('claim.namespace = :namespace', {
+      qb.andWhere('claim.claimType = :namespace', {
```
JGiter, line 175: Are you sure this is correct? I think claimType is not the same as namespace

jrhender: I think this might be right. I think we definitely have been using `claimType` as @dawidgil described. It looks like we have both `namespace` and `claimType` on `RoleClaim` entity. 

@JGiter I think `namespace` was a rename from `parentNamespace`. Do you remember the context of this change? https://github.com/energywebfoundation/ssi-hub/commit/90a7b09e0b6799eefd47431a6369def7eddfe604

Harasz: Huh, I think @JGiter is right. The problem is in ICL when we are passing `claimType` as a `namespace`.

JGiter: @jrhender Initially this parameter was called `namespace` or `parentNamespace` in different parts. We renamed it to be consistent

PR #373, opened 2022-04-12
ci: remove node-prune script

Harasz:


jrhender (commented):
I think it would be good to mention the deprecation that is prompting this removal in the commit description https://github.com/goreleaser/godownloader/issues/207

It would be hard otherwise for someone looking at the commit list to know why this commit was made

Harasz:
@jrhender added

jrhender (approved):
Thanks, looks good

PR #381, opened 2022-04-27
Update ENS_URL value

royki:
Hi @jrhender ,
`ENS_URL` value is using fast rpc url. Let me know if it uses archive rpc. I'll update then


michalziobro (approved):

jrhender (approved):
Fast RPC is okay. This change looks good.

PR #392, opened 2022-05-17
release 2.5.0

JGiter:


Harasz (changes requested):
It took commits that were committed in Jun 8, 2021

JGiter:
> It took commits that were committed in Jun 8, 2021

Ah, right. I guess this is because last tag was added in 2020. Will fix changelogs this time manually. Thanks for noticing

PR #398, opened 2022-05-23
fix: add IPFS CID checking

Harasz:


src/modules/did/did.service.spec.ts
```diff
@@ -189,6 +189,29 @@ describe('DidDocumentService', () => {
     expect(refreshedDID).resolves.toEqual(cachedDoc.id);
   });
 
+  it('should check valid cid', async () => {
```
jrhender, line 192:
```suggestion
  it('should check valid IPFS cid', async () => {
```

jrhender (approved):
Looks good to me, thanks @Harasz

PR #400, opened 2022-05-23
feat: avoid report common error to Sentry

Harasz:


JGiter (approved):

src/modules/sentry/sentry.service.ts
```diff
@@ -29,7 +30,28 @@ export class SentryService implements OnModuleDestroy, OnApplicationShutdown {
...
+    }
+
+    if (
+      exception instanceof QueryFailedError &&
+      exception.message.search(
+        'duplicate key value violates unique constraint'
+      ) !== -1
+    ) {
+      return true;
+    }
```
jrhender, line 44: I think we should include a comment explaining why we are excluding this error.
Please add/correct my explanation based on your understanding
```suggestion
    // We are excluding this exception from Sentry logs because this exception is expected behaviour from DB transaction validations
    // We have noticed that it is usually occurring when...
    if (
      exception instanceof QueryFailedError &&
      exception.message.search(
        'duplicate key value violates unique constraint'
      ) !== -1
    ) {
      return true;
    }
```

src/modules/sentry/sentry.service.ts
```diff
@@ -29,7 +30,28 @@ export class SentryService implements OnModuleDestroy, OnApplicationShutdown {
...
+
+    if (
+      exception instanceof Error &&
+      (exception.message === 'Invalid DID' ||
+        exception.message === 'Invalid did provided')
+    ) {
+      return true;
+    }
```
jrhender, line 52: Same rationale as the first comment.
```suggestion
    // This error is excluded because...
    if (
      exception instanceof Error &&
      (exception.message === 'Invalid DID' ||
        exception.message === 'Invalid did provided')
    ) {
      return true;
    }
```

src/modules/sentry/sentry.service.ts
```diff
@@ -29,7 +30,28 @@ export class SentryService implements OnModuleDestroy, OnApplicationShutdown {
...
+        'duplicate key value violates unique constraint'
+      ) !== -1
+    ) {
+      return true;
+    }
+
+    if (
+      exception instanceof Error &&
+      (exception.message === 'Invalid DID' ||
+        exception.message === 'Invalid did provided')
```
jrhender, line 49: Would it be possible to use string constants here? I'm assuming we have this string literal elsewhere in the app and so this is a duplication of the string literal

jrhender (commented):
Maybe instead of having excluded errors in Sentry, we should change the log level to `warn` instead of `error` for these case?

src/modules/sentry/sentry.service.ts
Harasz: Unfortunately errors are coming from `ew-did-registry` and right now there is now way to import them bc their hardcoded

Harasz:
@jrhender those errors are coming from different places, so it's easier to suppress them, then looking for all possible places, especially when it's catch by `try-catch` like many other errors

PR #401, opened 2022-05-24
fix(assets): fix parsing assets logs

Harasz:
The problem was that `at` is a built-in function for arrays in node 16 (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at#browser_compatibility) so it's shadow out property in parsed logs. We have only one BigNumber object in parsed logs which is our old `at` value, so that's why this fix looks like this.

### DIFF
--- a/src/modules/assets/assets.service.ts
+++ b/src/modules/assets/assets.service.ts
@@ -2,6 +2,7 @@ import { Injectable } from '@nestjs/common';
 import { ConfigService } from '@nestjs/config';
 import { InjectRepository } from '@nestjs/typeorm';
 import { Repository, Brackets } from 'typeorm';
+import { LogDescription } from 'ethers/lib/utils';
 import { Provider } from '../../common/provider';
 import { IdentityManager } from '../../ethers/IdentityManager';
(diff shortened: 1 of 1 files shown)


src/modules/assets/assets.service.ts
```diff
@@ -351,7 +351,9 @@ export class AssetsService {
       return {
         identity: parsedLog.args.identity,
         owner: parsedLog.args.owner,
-        at: parsedLog.args.at,
+        at:
```
JGiter, line 354: This should work too. Another option is https://github.com/energywebfoundation/ew-credentials/blob/0f3796bce1a7c00ddc27a979cb94581aa5984eea/packages/credential-governance/src/domain-hierarchy.ts#L241

JGiter (approved):

src/modules/assets/assets.service.ts
Harasz: I tried your solution, but the same issue.

Harasz:
@JGiter Take a look one more time. I found other places with similar error

JGiter (commented):
Do you know why `at` not being parsed now?

Harasz:
@JGiter same issue

PR #403, opened 2022-05-25
feat: status list module

Harasz:
Add StatusList2021 Module and features
Upgrade `typeorm` to latest version (breaking changes to API)
Add support for role definition v2


src/modules/status-list/status-list.controller.ts
```diff
@@ -0,0 +1,101 @@
...
+
+@Controller({ path: STATUS_LIST_MODULE_PATH, version: '1' })
+@ApiTags('status-list')
+export class StatusListController {
+  constructor(
+    private readonly statusListService: StatusListService,
+    private readonly roleService: RoleService
+  ) {}
+
+  @Post('/create-entry')
```
jrhender, line 32: As it is a `POST` already, I think that we don't need to specify `create`.
```suggestion
  @Post('/entries')
```

src/modules/status-list/status-list.service.ts
```diff
@@ -0,0 +1,113 @@
...
+      where: {
+        credentialId: credential.id,
+      },
+    });
+
+    if (statusList) {
+      throw new ConflictException('Credential already revoked');
+    }
+
+    const list = await StatusList.createList({ length: 16000 });
```
jrhender, line 58: Why length `16000` here? I think we only need length `1` because we going to do a 1 to 1 mapping from VC to StatusList2021Credential
```suggestion
    const list = await StatusList.createList({ length: 1 });
```

src/modules/status-list/status-list.controller.ts
```diff
@@ -0,0 +1,101 @@
...
+  async createEntry(
+    @User() currentUser: string,
+    @Body() { credential }: CreateEntryInputDto
+  ): Promise<CredentialWithStatusDto> {
+    const namespace = credential.credentialSubject.role.namespace;
+    const issuer = credential.issuer;
+
+    if (issuer !== currentUser) {
+      throw new UnauthorizedException(
+        'You are not an issuer of given credential'
```
jrhender, line 43:
```suggestion
        'Requester is not the issuer of given credential'
```

src/modules/status-list/status-list.controller.ts
```diff
@@ -0,0 +1,101 @@
...
+    await this.roleService.verifyEnrolmentIssuer({
+      issuerDID: issuer,
+      claimType: namespace,
+    });
+
+    return await this.statusListService.createEntry(credential);
+  }
+
+  @Post('/register-revoke')
+  async registerRevoke(
```
jrhender, line 56: We're straying from VC-API due to our client-side keys, but I think that this path is a bit more elegant and communicates that we are keeping VC-API in mind https://w3c-ccg.github.io/vc-api/#update-status
```suggestion
  @Post('/credentials/update/register')
  async registerStatusUpdate(
```

src/modules/status-list/status-list.controller.ts
```diff
@@ -0,0 +1,101 @@
...
+    // Check if current user is authorized to revoke given credential
+
+    return await this.statusListService.createStatusListCredential(
+      verifiableCredential,
+      currentUser
+    );
+  }
+
+  @Post('/sign-revoke')
+  async signRevoke(
```
jrhender, line 72: In the same spirit as my previous comment. I also think that `signRevoke` is a bit misleading here as the signature is happening client-side
```suggestion
  @Post('/credentials/update/finalize')
  async persistUpdate(
```

jrhender (commented):
Still reviewing, but here are some initial comments

src/modules/status-list/status-list.service.ts
Harasz: > The encodedList property of the credential [subject](https://w3c-ccg.github.io/vc-status-list-2021/#dfn-subjects) MUST be the GZIP-compressed [[RFC1952](https://w3c-ccg.github.io/vc-status-list-2021/#bib-rfc1952)], base-64 encoded [[RFC4648](https://w3c-ccg.github.io/vc-status-list-2021/#bib-rfc4648)] bitstring values for the associated range of [verifiable credential](https://w3c-ccg.github.io/vc-status-list-2021/#dfn-verifiable-credentials) status values. The uncompressed bitstring MUST be at least **16KB** in size.

see : https://w3c-ccg.github.io/vc-status-list-2021/#statuslist2021credential

src/modules/status-list/status-list.controller.ts
jrhender: Actually, maybe this is better
```suggestion
  @Post('/credentials/update/initiate')
  async initiateStatusUpdate(
```

src/modules/status-list/status-list.service.ts
jrhender: Hmm, I think this 16KB minimum is for privacy reasons that don't really apply in our case. Or can you think of a good reason for the 16KB minimum?

Harasz: I think just the privacy reasons.

src/modules/status-list/status-list.service.ts
```diff
@@ -0,0 +1,115 @@
...
+    private readonly statusListRepository: Repository<StatusListCredential>
+  ) {}
+
+  async createEntry(
+    credential: CredentialDto
+  ): Promise<CredentialWithStatusDto> {
+    const statusListCredential = new URL(
+      `${STATUS_LIST_MODULE_PATH}/${credential.id}`,
+      this.configService.get('STATUS_LIST_DOMAIN')
+    );
```
jrhender, line 32: I wonder if we should persist the entries we create and check if we have already created an entry with the same `credential.id`. Ideally the client should be creating unique ids for each credential but if they don't I think it could cause a problem where we would be giving out two credentials with the same status list entry and then subsequent updates to the statusListCredential could overwrite eachother.

jrhender (commented):
Added comment about possible conflicting `credential.id`

src/modules/status-list/status-list.service.ts
Harasz: Yeah, maybe this is a good idea. I will add it

src/modules/status-list/status-list.service.ts
```diff
@@ -0,0 +1,139 @@
...
+    private readonly configService: ConfigService,
+    @InjectRepository(StatusListCredential)
+    private readonly statusListRepository: Repository<StatusListCredential>
+  ) {}
+
+  async createEntry(
+    credential: CredentialDto
+  ): Promise<CredentialWithStatusDto> {
+    if (await this.fetchStatusList(credential.id)) {
+      throw new ConflictException('Credential already revoked');
```
JGiter, line 31:
```suggestion
      throw new ConflictException('Revocation status for this credential already created');
```

src/modules/status-list/status-list.service.ts
```diff
@@ -0,0 +1,139 @@
...
+      type: 'StatusList2021Entry',
+      statusPurpose: 'revocation',
+      statusListIndex: '1', // because we are using one-to-one mapping
+      statusListCredential: statusListCredential.href,
+    };
+
+    return { ...credential, credentialStatus };
+  }
+
+  async createStatusListCredential(
```
JGiter, line 50:
```suggestion
  async revokeCredential(
```

JGiter (commented):
From my point of view `entry` used in API relates to implementation details. `revocation status` would be more clear, like instead of `createEntry` `addRevocationStatus` or may be `makeCredentialRevocable`

src/modules/status-list/status-list.service.ts
```diff
@@ -0,0 +1,139 @@
...
+        type: 'StatusList2021',
+        statusPurpose: 'revocation',
+        encodedList,
+      },
+    };
+
+    return statusListCredential;
+  }
+
+  async addSignedStatusList(
```
JGiter, line 89: Could you add description to this method?

JGiter (commented):
There is no need to add `Dto` suffix to classes when there is no regular (non-Dto) classes like for CredentialDto
As introduced types will be used also in client side we should put them in @ew-did-registry/credential-interfaces and try to reuse existing defintions

src/modules/status-list/status-list.service.ts
Harasz: Not only it was created but it was actually revoked. We store status just for revoked credentials.

src/modules/status-list/status-list.service.ts
Harasz: We can't call it `revokeCredential` bc at this point credential is not revoked. Here we are just preparing object to sign by the revoker. Actual revocation is done in `addSignedStatusList`

src/modules/status-list/status-list.service.ts
Harasz: Yes, I will add it

Harasz:
From SSI-Hub perspective all those classes are Dtos. Even if we move some of them to `@ew-did-registry/credential-interfaces` I will need to add `ApiProperty` decorators or decorators from class-validator. @JGiter

src/modules/status-list/status-list.service.ts
```diff
@@ -0,0 +1,168 @@
...
+   * @param {VerifiableCredentialDto} credential
+   * @param {String} currentUser
+   * @return StatusList2021Credential without proof
+   */
+  async createStatusListCredential(
+    credential: VerifiableCredentialDto,
+    currentUser: string
+  ): Promise<StatusListCredentialDto> {
+    if (await this.fetchStatusList(credential.id)) {
+      throw new ConflictException('Credential already revoked');
```
JGiter, line 75:
```suggestion
      throw new ConflictException('Credential already exist');
```

JGiter (approved):

src/modules/status-list/status-list.service.ts
Harasz: Why `exists` it will always exists since we are passing verifiable credential that we want to revoke. Here is the revocation check

JGiter: I thought that this error is thrown when status list credential already exists. Probably more precise message is `Status list has already been created`?

JGiter (approved):

src/modules/status-list/status-list.service.ts
Harasz: If status list has been already created that mens credential is revoked

src/common/from-hex-did.ts
```diff
@@ -0,0 +1,16 @@
+export const fromHexDID = (did: string) => {
+  const didParts = did.split(':');
+
+  if (didParts.length !== 4) {
+    return did;
+  }
+
```
jrhender, line 7: I think this conversation is only relevant for `did:ethr` DIDs as far as I know. I think it could fail for other DID methods, so it's good to have a check for the method here, IMO.
```suggestion

if (didParts[1] !== 'ethr') {
  return did
}

```

src/common/to-hex-did.ts
```diff
@@ -0,0 +1,18 @@
+export const toHexDID = (did: string) => {
```
jrhender, line 1:
```suggestion
/*
 * Converts a `did:ethr` DID with a "chain name" chain id to its equivalent form with a hex chain id.
 * See https://github.com/decentralized-identity/ethr-did-resolver#multi-network-configuration for background.
 * Only works for `did:ethr` DIDs for the chain for which the library has a current connection.
 */
export const toHexEthrDID = (did: string) => {
```

src/modules/role/role.dto.ts
```diff
@@ -121,6 +152,9 @@ export class RoleDefinitionDTO implements IRoleDefinition {
   @IsDefined()
   issuer: IssuerDTO;
 
+  @IsDefined()
```
jrhender, line 155: Do you need `ValidatedNested` here too? My current understanding is that it is necessary, but I could be wrong. https://github.com/typestack/class-validator#validating-nested-objects

If it is needed, I think we should add it to the other properties of this class as well
```suggestion
  @IsDefined()
  @ValidatedNested()
  @Type(() => RevokerDTO)
```

jrhender (commented):
Still going through the PR, but some initial comments

devops/sandbox/values.yaml
```diff
@@ -89,6 +89,7 @@ iam-cache-server-helm:
     JWT_REFRESH_TOKEN_EXPIRES_IN: 1h
     JWT_ACCESS_TOKEN_NAME: token
     JWT_REFRESH_TOKEN_NAME: refreshToken
+    STATUS_LIST_DOMAIN: https://ics-dev.energyweb.org/v1
```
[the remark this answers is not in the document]
Harasz: I took the value from `STRATEGY_CACHE_SERVER`, should I change it also? And it's a sandbox env

src/common/from-hex-did.ts
Harasz: Not only it must be `ethr` but also DID with chain id

devops/sandbox/values.yaml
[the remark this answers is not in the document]
jrhender: Ah, I didn't realize it was `sandbox`. Maybe we just leave it then.

src/common/from-hex-did.ts
jrhender: Yea, I would keep the `didParts.length !== 4`. I just think if `didParts.length !== 4` is true but `didParts[1] !== 'ethr'` is false, then we probably don't want to process

src/common/to-hex-did.ts
```diff
@@ -0,0 +1,23 @@
+/*
+ * Converts a `did:ethr` DID with a "chain name" chain id to its equivalent form with a hex chain id.
+ * See https://github.com/decentralized-identity/ethr-did-resolver#multi-network-configuration for background.
+ * Only works for `did:ethr` DIDs for the chain for which the library has a current connection.
+ */
+export const toHexEthrDID = (did: string) => {
+  const didParts = did.split(':');
+
+  if (didParts.length !== 4) {
```
jrhender, line 9:
```suggestion
  if (didParts.length !== 4 || didParts[1] !== 'ethr') {
```

src/modules/status-list/status-list.service.ts
```diff
@@ -0,0 +1,169 @@
...
+   *
+   * @param {CredentialDto} credential
+   * @return credential with StatusList2021Entry
+   */
+  async addStatusListEntry(
+    credential: CredentialDto
+  ): Promise<CredentialWithStatusDto> {
+    if (await this.fetchStatusList(credential.id)) {
+      throw new ConflictException('Credential already revoked');
+    }
```
jrhender, line 44: I'm not sure this behaviour is desired. I'm imaging an edge case where an issuer has called `POST /entries`, but fails to actually issue the credential for some reason (maybe their browser crashes for example). Then if they try to call `POST /entries` again for the same `credential.id`, the request would fail.
Do you think it would reasonable if instead, the behaviour was that is just populates the `credentialStatus` with the existing entry if it exists?
```suggestion
    const existingEntry = await this.fetchStatusList(credential.id);
    if (existingEntry) {
      return { ...credential, credentialStatus: existingEntry }
    }
```

jrhender (commented):
Some further comments

src/modules/status-list/status-list.service.ts
Harasz: Exactly, we persisting only StatusList2021 Credential during revocation, so when revoker send signed status list credential.

jrhender (changes requested):
Some comments to discuss further

src/modules/status-list/status-list.controller.ts
```diff
@@ -0,0 +1,137 @@
...
+  @ApiBadRequestResponse()
+  async persistUpdate(
+    @User() currentUser: string,
+    @Body() { statusListCredential, revokedCredential }: SignRevokeInputDto
+  ): Promise<StatusListVerifiableCredentialDto> {
+    const issuer = fromHexEthrDID(statusListCredential.issuer);
+
+    await this.roleService.verifyRevoker({
+      revokerDID: currentUser,
+      claimType: revokedCredential.credentialSubject.role.namespace,
```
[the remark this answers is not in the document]
Harasz: Okay, make sens. Credential proof is checked in service

src/modules/status-list/status-list.controller.ts
```diff
@@ -0,0 +1,137 @@
...
+    });
+
+    if (revokedCredential.id !== statusListCredential.credentialSubject.id) {
+      throw new BadRequestException(
+        'Revoked credential does not match status list credential'
+      );
+    }
+
+    if (issuer !== currentUser) {
```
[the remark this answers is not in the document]
Harasz: I removed this check based on this comment https://github.com/energywebfoundation/ssi-hub/pull/403/files/d857b478f851e56bcf4bdb3b1cf573558ee4005f#r886828620

src/common/from-hex-did.ts
```diff
@@ -0,0 +1,26 @@
+/*
+ * Converts a `did:ethr` DID with a hex chain id to its equivalent form with a "chain name" chain id.
+ * See https://github.com/decentralized-identity/ethr-did-resolver#multi-network-configuration for background.
+ * Only works for `did:ethr` DIDs for the chain for which the library has a current connection.
+ */
+export const fromHexEthrDID = (did: string) => {
```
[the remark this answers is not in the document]
Harasz: I like that

src/modules/status-list/status-list.service.ts
```diff
@@ -0,0 +1,204 @@
...
+
+    const statusListCredential = new URL(
+      `${STATUS_LIST_MODULE_PATH}/${credential.id}`,
+      this.configService.get('STATUS_LIST_DOMAIN')
+    );
+
+    const newStatusList = StatusListCredential.create({
+      credentialId: credential.id,
+      namespace: credential.credentialSubject.role.namespace,
+    });
```
jrhender, line 54: To me, it would make more sense to use the `statusListCredential.id`, rather than the VC's id here. This would reduce the changes and confusion that would be necessary if the implementation changed to a "many VC to one status-list" relationship.
I guess a downside might be that we would need to use UUIDs for `StatusListCredential`, but I think this is worth it.
```suggestion
    const statusListCredentialId = uuid();
    const statusListCredentialUrl = new URL(
      `${STATUS_LIST_MODULE_PATH}/${statusListCredentialId}`,
      this.configService.get('STATUS_LIST_DOMAIN')
    );

    const newStatusList = StatusListCredential.create({
      id: statusListCredentialId,
      credentialId: credential.id,
      namespace: credential.credentialSubject.role.namespace,
    });
```

src/modules/status-list/entities/status-list-credential.entity.ts
```diff
@@ -0,0 +1,23 @@
...
+    const entity = new StatusListCredential();
+    Object.assign(entity, data);
+    return entity;
+  }
+
+  @PrimaryGeneratedColumn('uuid')
+  id: string;
+
+  @Column({ unique: true })
+  credentialId: string;
```
jrhender, line 16: I think eventually, if we had a "many credential to one status-list" situation, then we would want the `credential` to be a separate entity with a relationship to StatusListCredential so that it could have this many-to-one relationship. Do you agree with this? I think we keep it as is for now to keep it simple, but maybe add a comment to capture our rationale and thoughts?
```suggestion
  /*
   * The id of the associated Credential which references this StatusListCredential as its credentialStatus.
   *
   * Eventually, if there is a "many credential to one status-list" relationship,
   * then this property could be replace with a separate entity with a relationship to StatusListCredential.
   * However, this property is currently included in StatusListCredential for simplicity of implementation.
   */
  @Column({ unique: true })
  credentialId: string;
```

src/modules/status-list/status-list.service.ts
```diff
@@ -0,0 +1,204 @@
...
+   *
+   * @param {CredentialDto} credential
+   * @return credential with StatusList2021Entry
+   */
+  async addStatusListEntry(
+    credential: CredentialDto
+  ): Promise<CredentialWithStatusDto> {
+    if (await this.isCredentialRevoked(credential.id)) {
+      throw new ConflictException('Credential already revoked');
+    }
```
jrhender, line 44: I think this behaviour is not desired. If I have a `Credential` that I want to make revocable but I lose the `CredentialWithStatus` for some reason, I don't see why our API couldn't look up the `StatusListCredential` associated with this `credential.id` and reconstruct the `credentialStatus` to return the `CredentialWithStatusDto`.

jrhender (commented):
Some further comments

src/modules/status-list/status-list.service.ts
```diff
@@ -0,0 +1,204 @@
...
+  ): Promise<StatusListCredentialDto> {
+    const statusList = await this.fetchStatusList(credential.id);
+
+    if (!statusList) {
+      throw new BadRequestException('Credential was not registered');
+    }
+
+    if (statusList.credentialStatus) {
+      throw new ConflictException('Credential already revoked');
+    }
```
jrhender, line 90: I don't think this code is necessary. If someone wants to redo a revocation for some reason, I don't see why it shouldn't be possible in principle.
```suggestion
```

src/modules/status-list/status-list.controller.ts
```diff
@@ -0,0 +1,145 @@
...
+    });
+
+    return await this.statusListService.addStatusListEntry(credential);
+  }
+
+  @Auth()
+  @Post('/credentials/status/initiate')
+  @ApiCreatedResponse({ type: StatusListCredentialDto })
+  @ApiForbiddenResponse({ description: 'User is not authorized to revoke.' })
+  @ApiConflictResponse({ description: 'Credential already revoked.' })
```
jrhender, line 75: I think the documentation at the service level is great, but I think some documentation at the controller level would be helpful too.
```suggestion
  /*
   * Returns the StatusList2021Credential with the bit associated with the provided VerifiableCredential marked 
   * as true.
   * The updated StatusList2021Credential is not signed and not persisted.
   * https://w3c-ccg.github.io/vc-status-list-2021/#statuslist2021credential
   *
   * The operation is conducted server-side (rather than client-side) to encourage
   * clients to register their updates with the server, as this maybe required in future versions.
   * (For example, for the implementation of a many-to-one credential to status-list relationship).
   */
  @Auth()
  @Post('/credentials/status/initiate')
  @ApiCreatedResponse({ type: StatusListCredentialDto })
  @ApiForbiddenResponse({ description: 'User is not authorized to revoke.' })
  @ApiConflictResponse({ description: 'Credential already revoked.' })
```

src/modules/status-list/status-list.service.ts
```diff
@@ -0,0 +1,204 @@
...
+      statusPurpose: 'revocation',
+      statusListIndex: '0', // because we are using one-to-one mapping
+      statusListCredential: statusListCredential.href,
+    };
+
+    return { ...credential, credentialStatus };
+  }
+
+  /**
+   * Create a new StatusList2021Credential object for rejection purposes of a given credential for revoker.
```
jrhender, line 70: IMO this is a slightly clearer comment
```suggestion
   * Return a StatusList2021Credential which is updated with the revocation of the provided VC
```

src/modules/status-list/status-list.service.ts
```diff
@@ -0,0 +1,204 @@
...
+   * Object is not signed and not saved to database.
+   * https://w3c-ccg.github.io/vc-status-list-2021/#statuslist2021credential
+   *
+   * @param {VerifiableCredentialDto} credential
+   * @param {String} currentUser
+   * @return StatusList2021Credential without proof
+   */
+  async createStatusListCredential(
```
jrhender, line 78: A more accurate name, IMO
```suggestion
  async markStatusListCredential(
```

src/modules/status-list/entities/status-list-credential.entity.ts
Harasz: I think we can simply use an array of strings here instead of string. What you think? All the credentials would point to the same namespace and the same status list. Only the credential id can be different.

src/modules/status-list/status-list.service.ts
Harasz: Okay, If we lose credential status then we might have a problem, Will fix that

src/modules/status-list/entities/status-list-credential.entity.ts
jrhender: Yes, that could make sense. But it is good that we have the `credentialId` being `unique` for now and I think that might be harder to enforce with an array?

src/modules/status-list/status-list.service.ts
```diff
@@ -0,0 +1,204 @@
...
+    const array = Uint8Array.from([1]);
+    const encodedList = Buffer.from(gzip(array)).toString('base64');
+
+    // https://w3c-ccg.github.io/vc-status-list-2021/#statuslist2021credential
+    const statusListCredential = {
+      '@context': [
+        'https://www.w3.org/2018/credentials/v1',
+        'https://w3id.org/vc/status-list/2021/v1',
+      ],
+      id: credential.credentialStatus.statusListCredential,
```
jrhender, line 107: This seems confusing and maybe wrong to me. An issue that I am imagining that that what if the passed in credential had a `credential.id` that we had registered, the `namespace` was correct, the proof verified, but somehow the `credential.credentialStatus.statusListCredential` was wrong. I'm not sure it's really a security issue, just an opportunity to get weird behaviour.

Maybe we instead we should:
1. use the whole URL as our `credential.id` in `addStatusListEntry`? Then we can just do `statusList.id` here.
2. Or maybe we should extract a function that creates the URL from the id and configuration and then we can call it here. The downside to that might be that if the configuration values ever change, it would break the existing statusLists
3. Or maybe we store the components of the URL in the `CredentialStatusList` entity, then we can reconstruct the `statusListCredential.id` from our persisted values while also being able to run a query like `SELECT * WHERE STATUS_LIST_DOMAIN='https://identity-cache-dev.energyweb.org'`.

I guess I'm leaning towards option 3 at the moment. What are your thoughts @Harasz ?

jrhender (commented):
Some additional comments. The biggest one is around the `id` of the statusListCredential

src/modules/status-list/entities/status-list-credential.entity.ts
Harasz: Maybe it's better to change it now for string array. It will be much easier for us in the future

src/modules/status-list/status-list.service.ts
Harasz: Why `mark` is better?

src/modules/status-list/entities/status-list-credential.entity.ts
jrhender: I'm still not sure it makes sense to use a string array. If we have an array of `credentialId`, then do you think we would need to keep track of the `entry` associated with a given `credentialId`? I think we probably do as it doesn't make sense to have an array that breaks if you have more than one entry. So I think instead of an array of strings, we will need an array of a different entity.

However, I don't think I would necessarily implement the relationship between `StatusListCredential` and this new entity (say it is called `CredentialWithStatus`) as a nested relationship (where `StatusListCredential` is the aggregate-root, to use DDD language) as I think it would be better to keep the consistency boundary smaller.

Maybe the best implementation approach is two-step:
1. Claim an entry in `statusListCredential` of a `role` namespace
2. Associate that entry with a `credentialId`
This could lead to statusListEntries which are not associated with a credential, but I think this would be okay

src/modules/status-list/status-list.service.ts
Harasz: I agree, options 3

src/modules/status-list/entities/status-list-credential.entity.ts
jrhender: I think if we were looking ahead to a "many to one" implementation, this is the approach I would take now https://github.com/energywebfoundation/ssi-hub/blob/status-list-2022-implementation/docs/revocation.md#class-diagrams. What do you think?

Harasz:
@jrhender first draft of new approach

src/modules/status-list/status-list.service.ts
```diff
@@ -35,53 +46,64 @@ export class StatusListService {
...
+        `${STATUS_LIST_MODULE_PATH}/${credential.id}`,
+        this.configService.get('STATUS_LIST_DOMAIN')
+      ).href;
+
+      const entry = await this.statusListEntryRepository.save(
+        StatusListEntry.create({
+          statusListIndex: '0', // because we are using one-to-one mapping
+          statusListCredential,
+        })
+      );
```
jrhender: To me, I like encapsulating logic in entities. So I would put this in `NamespaceRevocation`. It has the "intelligence" on how to get an entry for a given namespace.
```
     const statusListCredential = new URL(
        `${STATUS_LIST_MODULE_PATH}/${credential.id}`,
        this.configService.get('STATUS_LIST_DOMAIN')
      ).href;
     // https://w3c-ccg.github.io/vc-status-list-2021/#statuslist2021entrya
     const entry: StatusList2021EntryDto = {
          type: 'StatusList2021Entry',
          statusPurpose: 'revocation',
          statusListIndex: '0', // because we are using one-to-one mapping
          statusListCredential,
     }
```
I also think that it makes sense to use a domain operation on `credentialWithStatus` as it is the aggregateRoot.
I agree with this [SO post](https://stackoverflow.com/a/1958668) that
> it is a common practice to have your repository only work with the entities that are Aggregate Roots, and this entity is responsible for initializing the other entities.

Even if we need to break aggregate encapsulation a bit to persist the entities, I think it's best to stick to it from a domain standpoint, as this give us more of the "onion architecture"

```suggestion
      const namespace = credential.credentialSubject.role.namespace;
      const namespaceRevocations = await this.namespaceRevocationsRepository.get(namehash(namespace));
      const domain = this.configService.get('STATUS_LIST_DOMAIN');
      const claimedEntry = namespaceRevocation.claimEntry(STATUS_LIST_MODULE_PATH, domain);
      await this.namespaceRevocationsRepository.save(namespaceRevocations);
      
      const credentialWithStatus = new credentialWithStatus(credential.id); // maybe better to do repo.create() here?
      credentialWithStatus.associateEntry(claimedEntry);
      await this.credentialWithStatusRepository.save(credentialWithStatus);
```

jrhender (commented):
Looks like great progress. I added some comments.

src/modules/status-list/entities/namespace-revocations.entity.ts
```diff
@@ -0,0 +1,16 @@
...
+@Entity()
+export class NamespaceRevocations {
+  static create(data: Omit<NamespaceRevocations, 'id'>): NamespaceRevocations {
+    const entity = new NamespaceRevocations();
+    Object.assign(entity, data);
+    return entity;
+  }
+
+  @PrimaryGeneratedColumn('uuid')
+  id: string;
```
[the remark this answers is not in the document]
Harasz: Make sense for me

src/modules/status-list/status-list.service.ts
Harasz: It all sounds good, but I see couple of drawbacks for our current situation. 
First of all, ssi hub is using different conventions, so it would make code inconsistent. Second problem is we are using here typeorm, so I would first need to find a solution to nicely pass a connection manager to entity class, which is not obvious for me right know and it would take time to figure it out. But I think we don’t have it right now. Probably at the end we will have to make a big refactor for many to many mapping anyway, so maybe then we can think more about it

jrhender: > so I would first need to find a solution to nicely pass a connection manager to entity class
I'm not really understanding why this is necessary. Let's discuss further if you have time

src/modules/status-list/status-list.controller.ts
```diff
@@ -0,0 +1,156 @@
...
+      throw new BadRequestException('Credential was not registered');
+    }
+
+    await Promise.all([
+      this.statusListService.verifyCredential(statusListCredential),
+      this.roleService.verifyRevoker({
+        revokerDID: issuer,
+        claimType: statusList.namespace,
+      }),
+    ]);
```
jrhender, line 133: I think this won't scale if we go to many-to-one credentials as the `revokedCredentialId = statusListCredential.credentialSubject.id` relationship won't hold. I think we should ask the client to provide the namespace and then search through the known status list ids for that namespace. See pseudocode here: https://github.com/energywebfoundation/ssi-hub/blob/status-list-2022-implementation/docs/revocation.md#update-entry-in-a-status-list

I think this is better than asking the client to provide the `verifiableCredential` again as they may want to perform a bulk update.

package.json
```diff
@@ -16,7 +16,7 @@
     "start:prod": "node dist/src/main",
     "lint": "eslint \"{src,apps,libs,test}/**/*.ts\" --fix",
     "test": "NODE_ENV=test jest -i \"/src/.*\\.spec\\.ts$\" --verbose --runInBand --force-exit",
-    "test:e2e": "run-s migration:test && jest -i \"/test/.*\\.e2e.spec.(ts|js)$\" --force-exit",
+    "test:e2e": "jest -i \"/test/.*\\.e2e.spec.(ts|js)$\" --force-exit",
```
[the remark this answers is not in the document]
Harasz: No no, we still need it

src/modules/claim/resolvers/credential.resolver.ts
```diff
@@ -0,0 +1,36 @@
...
+  constructor(private readonly claimService: ClaimService) {}
+
+  async getCredential(
+    subject: string,
+    claimType: string
+  ): Promise<IVerifiableCredential> {
+    const claim = await this.claimService.getByClaimType({
+      subject,
+      claimType,
+    });
```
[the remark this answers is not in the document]
Harasz: I'd suggest to update it in another task. We want to add StatusList module to SSI-Hub in this PR. Currently we have more then that and its starting to be challenging to truck all changes. So lets make that in other task/PR. Could be right away after this one.

src/modules/status-list/entities/credential-with-status.entity.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  @OneToOne(() => StatusListEntry, { cascade: true })
+  @JoinColumn()
+  entry: StatusListEntry;
+
+  public associateEntry(entry: StatusListEntry) {
+    this.entry = { ...entry, id: undefined };
+  }
+
+  public getCredentialSubject() {
```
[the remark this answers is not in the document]
Harasz: ofc, I meant  `getCredentialStatus` :face_with_head_bandage:

src/modules/status-list/status-list.service.ts
```diff
@@ -0,0 +1,299 @@
...
+      this.configService.get('STATUS_LIST_DOMAIN'),
+      credentialId
+    );
+
+    const credentialWithStatus = CredentialWithStatus.create({
+      id: credentialId,
+      namespace,
+    });
+
+    credentialWithStatus.associateEntry(entry);
```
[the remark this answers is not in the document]
Harasz: What if we add a check to `associateEntry` if entry exists then throw an error? And another way to block update at database level https://dba.stackexchange.com/questions/235003/how-to-prevent-update-of-existing-rows-record-but-allow-insert-using-postgresql. What you think?

src/modules/status-list/status-list.service.ts
```diff
@@ -0,0 +1,299 @@
...
+
+    await this.credentialWithStatusRepository.manager.transaction(
+      async (transactionalEntityManager) => {
+        await transactionalEntityManager.save(credentialWithStatus);
+        await transactionalEntityManager.save(namespaceStatusLists);
```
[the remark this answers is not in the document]
Harasz: Yes, it does and only one site of relations can be cascade

src/modules/status-list/status-list.service.ts
```diff
@@ -0,0 +1,299 @@
...
+
+  /**
+   * Check if credential is revoked. Revoked credential has StatusList2021Credential.
+   *
+   * @param {String} credentialId verifiable credential id
+   * @return true if credential is revoked
+   */
+  async isCredentialRevoked(credentialId: string): Promise<boolean> {
+    const credential = await this.getCredential(credentialId);
+    if (!credential) return false;
```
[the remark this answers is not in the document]
Harasz: So we should be default expect that credential is revoked?

src/modules/status-list/status-list.service.ts
```diff
@@ -0,0 +1,299 @@
...
+    }
+  }
+
+  /**
+   * Check if credential is revoked. Revoked credential has StatusList2021Credential.
+   *
+   * @param {String} credentialId verifiable credential id
+   * @return true if credential is revoked
+   */
+  async isCredentialRevoked(credentialId: string): Promise<boolean> {
```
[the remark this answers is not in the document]
Harasz: Ohh, I see that we are not using this method anymore. I was using it in previous approach. I will remove it

src/modules/status-list/status-list.controller.ts
```diff
@@ -0,0 +1,173 @@
...
+  ): Promise<Response> {
+    const credential = await this.statusListService.getCredential(credentialId);
+    const statusList = await this.statusListService.getStatusList(
+      credential?.entry?.statusListCredential
+    );
+
+    return response
+      .status(statusList?.vc ? HttpStatus.OK : HttpStatus.NO_CONTENT)
+      .send(statusList?.vc || null);
+  }
```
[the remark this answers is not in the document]
Harasz: Currently, every status list id is pointing here. So the `statusListCredential` is the link to this controller method. In the future it will change from `credentialId` to `statusListId` which will be `uuid` i believe. Maybe it's better to already have it like this :thinking:

jrhender (approved):
Looks good, thanks for all of your hard work on this @Harasz

PR #413, opened 2022-06-14
fix: fix verifying enrolment issuer

Harasz:


JGiter (commented):
What do you think of moving this to `DID` class https://github.com/energywebfoundation/ssi-hub/blob/develop/src/modules/did/did.types.ts and adding DIDPipe to the rest of the methods?

nichonien (approved):

Harasz:
@JGiter you mean add did comprehension?

PR #425, opened 2022-06-27
fix: multiple claim requests

Harasz:


src/modules/claim/claim.controller.ts
```diff
@@ -182,14 +181,15 @@ export class ClaimController {
...
+
     if (!result.isSuccessful) {
       throw new HttpException(result.details, HttpStatus.BAD_REQUEST);
     }
 
     await this.nats.publishForDids(
       ClaimEventType.REQUEST_CREDENTIALS,
       NATS_EXCHANGE_TOPIC,
-      [claimData.claimIssuer],
+      [claimData.claimIssuer?.[0]],
```
jrhender, line 192: `claimIssuer` is an array? Seems strange as there can only be one issuer identifier

jrhender (commented):
Why the change to an array of issuers?

src/modules/claim/claim.controller.ts
Harasz: it always be an array, i just find out that SSI-Hub has outdated interfaces so i fixed it here also
https://github.com/energywebfoundation/iam-client-lib/blob/90b74b193217ee3b4d186774a182dd196a549e16/src/modules/messaging/messaging.types.ts#L11

jrhender: Hmm, I think the issue is actually the `@IsArray()` on the controller. In `NewClaimIssueDTO` and `ClaimRejectionDTO`, I think there should only be one `claimIssuer`. I don't see how having an array makes sense there.

Harasz: I did not add it, it always be there. I just adjusted interfaces to implementation. We can create a task to make `claimIssuer` a simple string and not an array. But we will need to make changes in ICL, and that would be the braking change for SSI-Hub (if we not introduce v2 controller)

jrhender: How do issuers get notified for requests then? Maybe we only support polling for requests and we don't have nats notifications for claim requests. Does that match the behaviour that you've observed?

Harasz: You mean kyc-app and file-issuer? Those app are not listening for specific address. If you are talking about SB, then issuer dont get notifications.

jrhender: Hmm, right. I guess my mental model of the expected behaviour was incorrect. I thought the issuer should get notifications, but it seems they don't.

.github/workflows/build-test.yml
```diff
@@ -85,6 +85,7 @@ jobs:
           ASSETS_MANAGER_ADDRESS: ${{secrets.ASSETS_MANAGER_ADDRESS}}
           DID_SYNC_MODE_FULL: false
           STATUS_LIST_DOMAIN: ${{secrets.STATUS_LIST_DOMAIN}}
+          NATS_ENVIRONMENT_NAME: dev
```
jrhender, line 88: Why this change in this PR?

Harasz: Because i saw that it has undefined value in nats subject, so I added it

PR #427, opened 2022-06-29
Task/ics 223 fix ens sync

Harasz:
Fix ens sync when role has no revoker property (RoleDefinition v1).
Refactor role ens sync.
Sort domain list before sync (I noticed much less errors).
Fix for status list: issuer in lowercase, expected from didkit library to verify VC correctly.

### DIFF
--- a/src/modules/ens/ens.service.ts
+++ b/src/modules/ens/ens.service.ts
@@ -194,7 +194,8 @@ export class EnsService implements OnModuleDestroy {
       domain: this._ROOT_DOMAIN,
       mode: 'ALL',
     });
-    return domains;
+    // Sorting to reduce "parent namespace does not exists" type errors
+    return domains.sort();
   }
 
   private async eventHandler({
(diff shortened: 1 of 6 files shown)


src/modules/ens/ens.service.ts
```diff
@@ -194,7 +194,7 @@ export class EnsService implements OnModuleDestroy {
       domain: this._ROOT_DOMAIN,
       mode: 'ALL',
     });
-    return domains;
+    return domains.sort();
```
jrhender, line 197: Why addition of `sort` here?

Harasz: I observed less errors like `parent organization does not exists`. but maybe i am wrong, i did not count it :smile:

src/modules/status-list/entities/status-list-credential.entity.ts
```diff
@@ -38,7 +38,7 @@ export class StatusListCredential {
         statusPurpose: 'revocation',
         encodedList,
       },
-      issuer: new DID(issuerDid).withHexChain(),
+      issuer: new DID(issuerDid).withHexChain().toLowerCase(),
```
jrhender, line 41: Why the `toLowerCase()` here? I think this could could convert the address portion of the `ethr` DID away from the checksum address which wouldn't be ideal https://coincodex.com/article/2078/ethereum-address-checksum-explained/

src/modules/ens/ens.service.ts
jrhender: Hmm, not a bad idea, maybe that makes sense. Maybe just add a comment that this is why we are sorting in this case.
```suggestion
    // Sorting to reduce "parent namespace does not exists" type error
    return domains.sort();
```

src/modules/status-list/entities/status-list-credential.entity.ts
Harasz: @dawidgil reported an error when for revoked role ICL `isRevoked` result with false because verifying credential failed with error `0xa028720bc0cc22d296dcd3a26e7e8aae73c9b6f3, expected 0xA028720Bc0cc22d296DCD3a26E7E8AAe73c9B6F3`. I check and for all VC in ICL we have issuer in lowecase.

src/modules/ens/ens.service.ts
Harasz: Ok, done

JGiter (approved):

src/modules/status-list/entities/status-list-credential.entity.ts
jrhender: In my opinion, the representation of the address that we should be striving in all cases is the checksum address. We can use `toLowerCase` for implementation of comparisons but I think in any persisted or external data, we should use checksum address.

In the case of `isRevoked` that you described, it sounds to me like it is the behaviour of `isRevoked` that should change, not the value used in the `StatusListCredential`. But I'm not 100% sure about this. How do you see it?

jrhender: > This requirements comes from didkit library. So we can create ticket in their board and ask about it

Ah you're saying that DIDKit expects the `issuer` to be lowercase when verifying proofs?

> Second thing is that we can't persist something in checksum format but send with lowecase format, because the proof value will be invalid.

Right, I was thinking we would persist and send both as checksum. but if DIDKit proof verification requires lowercase, then we might need to send as lowercase

Harasz: Exactly,  `didkit` proof verification requires lowercase. I checked in ICL and all VCs have lowercase issuer as well

jrhender: Ok, thanks for walking me through it. I think it would make sense to add a comment why this is then.
```suggestion
      // toLowerCase used because DIDKit proof verification requires lowercase
      issuer: new DID(issuerDid).withHexChain().toLowerCase(),
```

jrhender (approved):

PR #428, opened 2022-06-30
fix: updating roles definition

Harasz:
`role` is the previous object in database and `data` is the new definition. It always save old definition during update.


jrhender (commented):
@Harasz Do you know why the lines you removed were there in the first place?

Harasz:
I don't know. Maybe @JGiter know more

jrhender (approved):

PR #432, opened 2022-07-06
build(deps): upgrade `passport-did-auth`

Harasz:


src/modules/claim/services/claim.service.ts
```diff
@@ -306,6 +306,10 @@ export class ClaimService {
     }
     const rolesByRevoker = allRolesByRevoker.map((r) => r.namespace);
 
+    if (!rolesByRevoker.length) {
```
JGiter, line 309: If rolesRevokers can't be undefined then we can return just roleRevokers

JGiter (approved):

src/modules/claim/services/claim.service.ts
Harasz: this function needs to return claims not roles

Harasz:
blocked by https://github.com/energywebfoundation/passport-did-auth/pull/304

PR #433, opened 2022-07-08
chore: update `typedoc` config

Harasz:


jrhender (approved):
Ah yes, this is wonderful, thank you

PR #441, opened 2022-07-20
feat: handle unexpected did service

Harasz:


src/modules/did/did.service.ts
```diff
@@ -413,22 +415,42 @@ export class DIDService implements OnModuleInit {
...
+          const { claimData, ...claimRest } = decodedData;
 
-        const { claimData, ...claimRest } = decodedData;
+          return {
+            serviceEndpoint,
+            ...rest,
+            ...claimData,
+            ...claimRest,
+          };
+        }
```
jrhender, line 435: Do you think it makes sense to put this `isJWT` branch in the `try/catch` as well?

src/modules/did/did.service.ts
```diff
@@ -413,22 +415,42 @@ export class DIDService implements OnModuleInit {
...
+              ...rest,
+              verifiableCredentials: data,
+            };
+          }
+          return {
+            serviceEndpoint,
+            ...rest,
+            ...data,
+          };
+        } catch {
```
jrhender, line 451: Do you think it makes sense to log at `debug` level here? It could get pretty verbose if we have lots of these, but it could be good to have to troubleshoot

jrhender (commented):
Looks pretty good, added a few comments

jrhender (approved):
looks good to me, thanks @Harasz

JGiter (approved):

PR #442, opened 2022-07-20
build(deps): update dependencies

Harasz:


jrhender (commented):
Are you sure there are no breaking changes in the `@nestjs/axios` update? There shouldn't be, but I've add axios break on minor version increase before I think

Harasz:
@jrhender release message doesn't mention about breaking changes, just deps update -> https://github.com/nestjs/axios/releases/tag/0.1.0

jrhender (approved):

PR #445, opened 2022-07-21
update ew-credentials

JGiter:


jrhender:
@JGiter Please take a look at the "Build and test" errors

src/modules/claim/resolvers/credential.resolver.ts
```diff
@@ -1,17 +1,31 @@
...
   constructor(private readonly claimService: ClaimService) {}
-
   async getCredential(
+    did: string,
+    namespace: string
+  ): Promise<
+    VerifiableCredential<RoleCredentialSubject> | RoleEIP191JWT | null
+  > {
+    const vc = await this.getVerifiableCredential(did, namespace);
+    const eip191Jwt = await this.getEIP191JWT(did, namespace);
```
Harasz, line 21: Maybe we can optimize it a little bit. We need vc or jwt, so why fetch both?

src/modules/claim/resolvers/credential.resolver.ts
```diff
@@ -21,15 +35,30 @@ export class RoleCredentialResolver implements CredentialResolver {
...
     const claim = await this.claimService.getByClaimType({
       subject,
       claimType,
     });
+    if (!claim) {
+      return null;
+    }
+
+    const { issuedToken } = claim;
+    const { claimData, signer, credentialStatus } = jsonwebtoken.decode(
```
Harasz, line 51: What if claim is only on-chain, so issuedToken is undefined?

Harasz (changes requested):

src/modules/claim/resolvers/credential.resolver.ts
JGiter: Yes, make sense

jrhender (approved):
This looks good to me

JGiter:
Could some admin merge? I unable because of audit failure

jrhender:
I think we also need @Harasz to review again as he requested changes

Harasz (approved):

PR #448, opened 2022-07-26
task/ICS-56 unable to query credential from browser

JGiter:


jrhender (commented):
I think the approach is good. Just need to fix the tests https://github.com/energywebfoundation/ssi-hub/runs/7518245724?check_suite_focus=true

src/common/url-decode.decorator.ts
```diff
@@ -0,0 +1,11 @@
+import { createParamDecorator, ExecutionContext } from '@nestjs/common';
+
+/**
+ * Decodes URL encoded request url
+ */
+export const UrlDecoded = createParamDecorator(
```
jrhender, line 6: "request url" is ambiguous between just the "original URL" that we are using or the whole URL with the domain, so better to be more specific
```suggestion
/**
 * Decodes URL encoded "originalUrl"
 * https://expressjs.com/en/api.html
 */
export const OriginalUrlDecoded = createParamDecorator(
```

jrhender (approved):
Looks good!

PR #449, opened 2022-07-27
test ArgoCD notification

JGiter:


src/modules/auth/login.controller.ts
```diff
@@ -42,6 +42,9 @@ export class LoginController {
   @Post('login')
   async login(@Req() req: Request, @Res() res: Response) {
     const origin = req.headers['origin'];
+    if (origin.includes('switchboard')) {
+      throw new Error('Switchboard is not allowed!');
+    }
```
jrhender, line 47: The objective is to trigger the health-checks to fail for ssi-hub, not break from a functional stand point (which is what I think this is).

jrhender (commented):
Could you change the commit message to `chore` instead of `fix` so that this commit doesn't end up in the changelog?

PR #451, opened 2022-07-27
chore: remove Error on startup

jrhender:
Had introduced to test Argo notifications


jrhender:
No sure why the documentation is changed here. Perhaps because I did a `git pull` before the PR. It looks like it the correct documentation but from this PR https://github.com/energywebfoundation/ssi-hub/pull/448
Any thoughts here @JGiter ? We can make another bug to investigate this

JGiter (approved):

PR #452, opened 2022-07-27
get credential status from swagger

JGiter:


jrhender:
@JGiter Does it work if you access the swagger while running locally?

JGiter:
> Does it work if you access the swagger while running locally?

Yes

jrhender (approved):
Thanks @JGiter !

PR #457, opened 2022-08-02
resource limits

JGiter:


jrhender (changes requested):
Instead of `develop`, let's see if we can try to go into sandbox

JGiter:
> Instead of `develop`, let's see if we can try to go into sandbox

Ok, I'll add sandbox environment and create `sandbox` branch off of develop. By the way are `staging`, `prod` and `sandbox` in `devops` folder used in `develop` branch?

jrhender (approved):
Let's try it!

PR #461, opened 2022-08-08
task/ICS-231 align resources

JGiter:


devops/dev/values.yaml
```diff
@@ -126,24 +126,20 @@ iam-cache-server-helm:
...
-    # requests:
-    #   cpu: 100m
-    #   memory: 128Mi
+  resources:
+    limits:
+      cpu: 750m
+      memory: 1Gi
+    requests:
+      cpu: 500m
+      memory: 900Mi
```
jrhender, line 135:
```suggestion
      # High memory request is due to DIDKit usage https://github.com/spruceid/didkit/issues/301
      # We anticipate memory usage to decrease once node v18 is used
      memory: 900Mi
```

jrhender (approved):
Looks good, but I think it is good to have a comment to explain why the high request

jrhender:
@JGiter I also think that the commit message could be improved here. The commit is doing more than adding replicas (which were there already I believe). The core change is the addition of the resource limits so I think this should be reflected in the commit message

JGiter:
You are right. I'll create several commits

jrhender (approved):
Looks good, thanks!

PR #466, opened 2022-08-12
update did-registry

JGiter:
Don not merge until IPFS re-configured


nichonien (approved):

jrhender (changes requested):
I think we should hold off on this until we get IPFS fix into `staging` as this change isn't critical IMO

src/modules/claim/resolvers/credential.resolver.ts
```diff
@@ -45,13 +47,42 @@ export class RoleCredentialResolver implements CredentialResolver {
...
+    });
+
+    return claims
+      .map(
+        (claim) =>
+          claim?.vp
+            ?.verifiableCredential?.[0] as VerifiableCredential<RoleCredentialSubject>
+      )
+      .filter(Boolean);
+  }
```
jrhender, line 79: Why are these two methods added? They don't seem to be used.

JGiter: they are part of CredentialResolver interface from ew-credentials

jrhender: Do you think it's better to get the credentials from the `claimService` or from the DID document / IPFS in this case?

JGiter: Perhaps yes, because it allow to verify issuer which prefer to not publish his credential

jrhender: We decided that issuers must publish their credentials. This is to not rely on a potentially centralized ssi-hub.

Therefore, I think this doesn't align then with our implementation from passport-did-auth and I think we should have alignment so that the behaviour is consistent across the stack. https://github.com/energywebfoundation/passport-did-auth/blob/2aa3e85ab316f474186c8eaf81f878c4c351155e/lib/RoleCredentialResolver.ts#L142

jrhender (changes requested):
I think we need to align on the verification approach and I think it should match what we are doing in passport-did-auth so that credential validation behaviour is consistent

src/modules/claim/resolvers/credential.resolver.ts
nichonien: @jrhender I think this is completely fine. Agreed that the credentials needs to be published so that it is verified through the IPFS resolution approach, but then there could be a case where user wants to fetch the cached credentials and interact with sis-hub by choice.

nichonien: IMO we can use these methods for simply fetching the credentials based on the request by holder ( for example displaying the holder's credential in switchboard ) and not use this for any verification.

JGiter: @jrhender If credentials should read from document can we use IpfsCredentialResolver from vc-verification for this? 
And also should issuer and revoker resolvers in ssi-hub also read from chain?

src/modules/did/did.service.ts
```diff
@@ -309,6 +309,16 @@ export class DIDService implements OnModuleInit {
     }
   }
 
+  public async resolveServiceEndpoints(did: string) {
```
jrhender, line 312: Could you add documentation and return type to this method please? Does this seem valuable from your perspective?

src/modules/claim/resolvers/credential.resolver.ts
```diff
@@ -1,69 +1,104 @@
...
+    const tokens = await this.didService.resolveServiceEndpoints(did);
     return (
-      (await this.getVerifiableCredential(did, namespace)) ||
-      (await this.getEIP191JWT(did, namespace))
+      this.serviceEndpointsToCredentials(tokens).find(
+        (cred) => cred.credentialSubject.role.namespace === namespace
+      ) ||
+      this.serviceEndpointsToEIP191(tokens).find(
+        (token) => token.payload.claimData.claimType === namespace
+      )
```
jrhender, line 30: Maybe it's good to have the optional chaining operator here in order to avoid an exception? I think it's best to have each individual method be as safe as possible.
```suggestion
      this.serviceEndpointsToCredentials(tokens).find(
        (cred) => cred?.credentialSubject?.role?.namespace === namespace
      ) ||
      this.serviceEndpointsToEIP191(tokens).find(
        (token) => token?.payload?.claimData?.claimType === namespace
      )
```

src/modules/claim/resolvers/credential.resolver.ts
```diff
@@ -1,69 +1,104 @@
...
   ): Promise<VerifiableCredential<RoleCredentialSubject> | null> {
-    const claim = await this.claimService.getByClaimType({
-      subject,
-      claimType,
-    });
-
-    return claim?.vp
-      ?.verifiableCredential?.[0] as VerifiableCredential<RoleCredentialSubject>;
+    return (await this.credentialsOf(subject)).find(
+      (token) => token.credentialSubject.role.namespace === namespace
```
jrhender, line 39: - Again, probably safer to use optional chaining
- I think credential is a clearer name
```suggestion
      (credential) => credential?.credentialSubject?.role?.namespace === namespace
```

src/modules/claim/resolvers/credential.resolver.ts
```diff
@@ -1,69 +1,104 @@
...
-  constructor(private readonly claimService: ClaimService) {}
+  constructor(private readonly didService: DIDService) {}
+
   async getCredential(
     did: string,
     namespace: string
   ): Promise<
     VerifiableCredential<RoleCredentialSubject> | RoleEIP191JWT | null
   > {
+    const tokens = await this.didService.resolveServiceEndpoints(did);
```
jrhender, line 23: IMO, this is a bit clearer of a name. didDocument service endpoints don't necessarily resolve to a `token` so it's a bit unintuitive that this is the name.
```suggestion
    const resolvedEndpoints = await this.didService.resolveServiceEndpoints(did);
```

jrhender (approved):
Some minor suggestion, but I think we could be good to proceed 👍

src/modules/did/did.service.ts
JGiter: Sure, I'll try to find eslint rule for this

src/modules/claim/resolvers/credential.resolver.ts
JGiter: I agree that credential needs this check, because `isVerifialbleCredential` checks only high-level properties https://github.com/energywebfoundation/ew-credentials/blob/19108ef86410e85aad71bc47d44ddf202ced716f/packages/vc-verification/src/utils/formatters.ts#L108, but for eip191 this is not needed, because whole chain is checked by `isEIP191Jwt` https://github.com/energywebfoundation/ew-credentials/blob/19108ef86410e85aad71bc47d44ddf202ced716f/packages/vc-verification/src/utils/formatters.ts#L97

PR #481, opened 2022-08-19
fix: add api query for asset id

whitneypurdum:


jrhender (approved):

jrhender:
Arguably, this is a `docs` change and not a `fix`, IMO, as it is not functional. But this is okay, not that important.

whitneypurdum:
> Arguably, this is a `docs` change and not a `fix`, IMO, as it is not functional. But this is okay, not that important.

Ah totally @jrhender . I just changed it. Thanks.

JGiter (approved):

PR #483, opened 2022-08-19
fix: send asset id as did to DID pipe

whitneypurdum:
- The DID pipe was sending ID and stripping away the chain info ('did'ethr...'). 
- (0xEa242C31FCcB83Aa1Bf3b29e4B4b276E85024aa8 from did:ethr:volta:0xEa242C31FCcB83Aa1Bf3b29e4B4b276E85024aa8)
- Asset IDs are saved in the database in this format, so it wasn't finding any records. 

ASSET_TRANSFERRED event for did:ethr:volta:0x921*** saved: {\**"assetId\":\"did:ethr:volta:0x921*****\",\"at\":PHONE_NUMBER,\"emittedBy\":\"did:ethr:volta:0x06B***\",\"timestamp\":\"DIGITS-08-19T13:37:50.000Z\",\"relatedTo\":\"did:ethr:volta:0x17b***\",\"type\":\"ASSET_TRANSFERRED\",\"id\":DIGITS}\n"

### DIFF
--- a/src/modules/assets/assets.controller.ts
+++ b/src/modules/assets/assets.controller.ts
@@ -66,7 +66,7 @@ export class AssetsController {
     enum: AssetHistoryEventType,
   })
   async getHistoryByAssetId(
-    @Param('id', DIDPipe) { id }: DID,
+    @Param('id', DIDPipe) { did: id }: DID,
     @Query('take', new DefaultValuePipe(10), ParseIntPipe) take?: number,
     @Query('skip', new DefaultValuePipe(0), ParseIntPipe) skip?: number,
     @Query('order') order?: Order,
--- a/src/modules/assets/assets.service.ts
+++ b/src/modules/assets/assets.service.ts
@@ -183,8 +183,8 @@ export class AssetsService {
       async (identity: string, owner: string, at: BigNumber) => {
         const atAsNumber = at.toNumber();
         const assetDto = await AssetDto.create({
-          id: getDIDFromAddress(identity),
-          owner: getDIDFromAddress(owner),
+          id: new DID(identity).did,
+          owner: new DID(owner).did,
           at: atAsNumber,
         });
 


src/modules/assets/assets.controller.ts
```diff
@@ -66,7 +66,7 @@ export class AssetsController {
     enum: AssetHistoryEventType,
   })
   async getHistoryByAssetId(
-    @Param('id', DIDPipe) { id }: DID,
+    @Param('id', DIDPipe) { did: id }: DID,
```
JGiter, line 69: :+1: 
Can we also replace `getDidFromAddress` `https://github.com/energywebfoundation/ssi-hub/blob/f7954c7d68791cc564c70470f22f45bf341ad560/src/modules/did/did.types.ts#L107` by new DID(address).did here https://github.com/energywebfoundation/ssi-hub/blob/f7954c7d68791cc564c70470f22f45bf341ad560/src/modules/assets/assets.service.ts#L180

whitneypurdum: @JGiter do you mean replace the getDIDFromAddress helper with the DID pipe?

src/modules/assets/assets.service.ts
```diff
@@ -183,8 +183,8 @@ export class AssetsService {
       async (identity: string, owner: string, at: BigNumber) => {
         const atAsNumber = at.toNumber();
         const assetDto = await AssetDto.create({
-          id: getDIDFromAddress(identity),
-          owner: getDIDFromAddress(owner),
+          id: new DID(identity).did,
```
JGiter, line 186: I think it would be safer to be consistent with this by using either `getDIDFromAddress` or `new DID`, but not both. It is not strictly related to this PR though, so we can change this in other PR

JGiter (approved):

PR #495, opened 2022-09-06
bug/ICS-239 chain response timeout

JGiter:


nichonien (approved):

src/common/provider.ts
```diff
@@ -1,10 +1,39 @@
...
   constructor(configService: ConfigService) {
     super(configService.get('ENS_URL'));
+    this.attempts = configService.get('CHAIN_REQUEST_ATTEMPTS');
+  }
+
+  /**
+   * Retries failed chain requests
+   * @override
+   */
+  public perform(method: string, params: any) {
```
jrhender, line 20: Is this mostly copied from the GitHub issue that you linked? If so, I think it is best to link to that issue here

src/common/provider.ts
```diff
@@ -1,10 +1,39 @@
...
+    let attempts = 0;
+    return poll(() => {
+      attempts++;
+      return super.perform(method, params).then(
+        (result) => {
+          return result;
+        },
+        (error) => {
+          if (attempts >= this.attempts) {
+            return error;
```
jrhender, line 30: Could we add a log statement here?

src/common/provider.ts
```diff
@@ -1,10 +1,43 @@
...
+      return super.perform(method, params).then(
+        (result) => {
+          return result;
+        },
+        (error) => {
+          if (attempts >= this.attempts) {
+            this.logger.warn(
+              `Failed to ${method} with params ${params} after ${this.attempts}`
+            );
+            return error;
```
artursudnik, line 34: This is going to resolve the promise returned by the `perform` method. Is this expected or should rejected promise be returned here?

JGiter: Ah, right. Error should be thrown in order to reject

artursudnik (approved):

src/common/provider.ts
```diff
@@ -1,10 +1,43 @@
...
 
 @Injectable()
-export class Provider extends JsonRpcProvider {
-  constructor(configService: ConfigService) {
+export class Provider extends providers.JsonRpcProvider {
+  private attempts: number;
+
+  constructor(configService: ConfigService, private logger: Logger) {
     super(configService.get('ENS_URL'));
+    this.attempts = configService.get('CHAIN_REQUEST_ATTEMPTS');
```
jrhender, line 14: Do you think we should have a default value for `this.attempts` if `CHAIN_REQUEST_ATTEMPTS` is not set? Otherwise, I think `if (attempts >= this.attempts)` could give us an endless loop of requests and errors

JGiter: I think that current defaul 0 attempts is reasonable default, which means that request will not be retry. Could you elaborate why this is causes endless loop? Rejected callback should not be rescheduled by `poll`. I'll try to test this
This defaul should be set explicitly though with explanation

jrhender (approved):

src/common/provider.ts
jrhender: > Could you elaborate why this is causes endless loop?
Maybe it doesn't, I'm not sure. I think with the default value for `CHAIN_REQUEST_ATTEMPTS` though, this is not a concern

PR #498, opened 2022-09-20
fix: add idleTimeoutMillis as 10 seconds

whitneypurdum:
- add config for idleTimeoutMillis - see bug https://energyweb.atlassian.net/browse/SWTCH-1497


src/db/config.ts
```diff
@@ -30,6 +30,7 @@ export const getDBConfig = (configService: ConfigService) => {
         'DB_CONNECTION_TIMEOUT',
         1000
       ),
+      idleTimeoutMillis: 10000,
```
jrhender, line 33: @whitneypurdum sorry, I was confused in our previous discussion. I think it's better to adjust `connectionTimeoutMillis`, rather than `idleTimeoutMillis`

PR #499, opened 2022-09-20
fix: increase connectionTimeoutMillis

whitneypurdum:
- increases connectionTimeoutMillis in attempt to address bug "error: timeout exceeded when trying to connect" https://energyweb.atlassian.net/browse/SWTCH-1497


jrhender (approved):
Looks good. IMO, it would better to put additional information in the commit body that describes why this change is being made. Maybe something like:
> Attempting to reduce frequency of "timeout exceeded when trying to connect" messages

PR #507, opened 2022-09-26
fix: log revoker error for debugging

whitneypurdum:
- verifyRevoker in ew-credentials throws several different errors. SSI Hub does not handle the reason at all, so there is no way of knowing why the verifyRevoker failed. This adds a catch so the logger can log the error description. I would add that we should add the reason to the throw message as well, but that is up for debate. At the very least, we can log the error.


jrhender (approved):
Looks good, thanks

nichonien (approved):

PR #508, opened 2022-09-27
rename `prod` Sentry environment to `production`

JGiter:


jrhender:
@JGiter the commit in the PR is strange. It seems to be a merge commit. Is this what you intended?

jrhender:
@JGiter rather than targeting a different environment in ssi-hub, I think we would be better to hide the `prod` environment and unhide the `production` environment in Sentry. In this way, we would keep all of the historical issue data associated with the `production` environment. So I think we should not merge this PR.
What do you think of this @JGiter @royki ?

royki:
> @JGiter rather than targeting a different environment in ssi-hub, I think we would be better to hide the `prod` environment and unhide the `production` environment in Sentry. In this way, we would keep all of the historical issue data associated with the `production` environment. So I think we should not merge this PR. What do you think of this @JGiter @royki ?

Hi @jrhender will do the the hide `prod` and unhide `production` .
Here is our conversation ->https://energywebfoundation.slack.com/archives/C01N703QADU/p1664267365142739?thread_ts=1664195378.277719&cid=C01N703QADU

JGiter:
> we would be better to hide the prod environment and unhide the production environment

I would prefer this. @royki Are we updating `SENTRY_ENV` from `prod` in ssi-hub, because its corresponds to `production` in Sentry?

royki:
> > we would be better to hide the prod environment and unhide the production environment
> 
> I would prefer this. @royki Are we updating `SENTRY_ENV` from `prod` in ssi-hub, because its corresponds to `production` in Sentry?

@JGiter 
I'm bit confused. However I'm ok to hide `prod` and unhide `production` in Sentry env and check.

PR #510, opened 2022-09-30
chore: bump ew-credentials to 2.2.1-alpha.303

whitneypurdum:
- bumps ew-credentials package so that we can get the version that throws the stack trace with the error
- updates interfaces so cache is set


src/modules/claim/resolvers/revoker.resolver.ts
```diff
@@ -1,8 +1,12 @@
 import { IRevokerDefinition } from '@energyweb/credential-governance';
```
whitneypurdum, line 1: @nichonien can you please let me know if this PR implements the cache correctly?

src/modules/claim/resolvers/credential.resolver.ts
```diff
@@ -104,4 +107,32 @@ export class RoleCredentialResolver implements CredentialResolver {
...
+  async getDIDDocument(
+    did: string,
+    didDocumentCache?: IDIDDocumentCache
+  ): Promise<any> {
+    const cachedDIDDocument = didDocumentCache?.getDIDDocument(did);
+    if (cachedDIDDocument) {
+      return cachedDIDDocument;
+    }
+    const resolvedDIDDocument =
+      await this.didService.getDIDDocumentFromUniversalResolver(did);
```
whitneypurdum, line 134: This method returns any instead of IDIDDocument

src/modules/claim/resolvers/credential.resolver.ts
```diff
@@ -104,4 +108,32 @@ export class RoleCredentialResolver implements CredentialResolver {
...
+  async getDIDDocument(
+    did: string,
+    didDocumentCache?: IDIDDocumentCache
+  ): Promise<IDIDDocument> {
+    const cachedDIDDocument = didDocumentCache?.getDIDDocument(did);
+    if (cachedDIDDocument) {
+      return cachedDIDDocument;
+    }
+    const resolvedDIDDocument =
+      await this.didService.getDIDDocumentFromUniversalResolver(did);
```
whitneypurdum, line 135: @nichonien can you let me know if this is implementing the getDIDDoc method correctly?

src/modules/claim/resolvers/revoker.resolver.ts
nichonien: @whitneypurdum The approach for caching has changed recently, now we just inject in methods itself - https://github.com/energywebfoundation/ew-credentials/blob/98b4d13b3c29a1e3767d1e059431e8892efdbe64/packages/vc-verification/src/resolver/issuer-resolver.ts#L46.

So it's best to get the latest version.

whitneypurdum:
As discussed with Dmitry and Ashish, we will use the latest available versions of ew-credentials AND ew-did-registry

whitneypurdum:
I will squash these commits on merging.

package.json
```diff
@@ -45,18 +45,18 @@
...
-    "@ew-did-registry/credentials-interface": "^0.7.1-alpha.795.0",
-    "@ew-did-registry/did-document": "^0.7.1-alpha.795.0",
-    "@ew-did-registry/did-ethr-resolver": "^0.7.1-alpha.795.0",
-    "@ew-did-registry/did-ipfs-store": "^0.7.1-alpha.795.0",
-    "@ew-did-registry/jwt": "^0.7.1-alpha.795.0",
-    "@ew-did-registry/keys": "^0.7.1-alpha.795.0",
-    "@ew-did-registry/proxyidentity": "^0.7.1-alpha.795.0",
+    "@ew-did-registry/claims": "0.8.1-alpha.933.0",
+    "@ew-did-registry/credentials-interface": "0.8.1-alpha.933.0",
+    "@ew-did-registry/did-document": "0.8.1-alpha.933.0",
```
nichonien, line 54: @JGiter I see you already have an open PR to upgrade `ew-did-registry` to v8.  Do you think we can upgrade that here ? Or should wait for the IPFS changes to be finalised?

src/modules/auth/login.strategy.ts
```diff
@@ -2,29 +2,42 @@ import { LoginStrategy } from 'passport-did-auth';
 import { Inject, Injectable } from '@nestjs/common';
 import { PassportStrategy } from '@nestjs/passport';
 import { ConfigService } from '@nestjs/config';
+import { LoginStrategyOptions } from 'passport-did-auth/dist/lib/LoginStrategy';
```
nichonien, line 5: @whitneypurdum I don't think if it would be a good idea to update `passport-did-auth` now, given the long login issue and also there are some changes in the pipeline reg the `iam-client-examples` update. So the behaviours and signature may change. @jrhender what do you think about this?

package.json
whitneypurdum: What version should I update to if not? This morning in the meeting we said if I used the latest Of ew-credentials, I should use the latest of ew-did-registry as well.

src/modules/auth/login.strategy.ts
whitneypurdum: https://www.npmjs.com/package/passport-did-auth

whitneypurdum: I just reverted it back to the original version!

package.json
JGiter: did-reg v0.8 has breaking changes. We should wait for merging PR integrates these changes in ssi-hub

PR #516, opened 2022-10-07
update ew-did-registry@v8

JGiter:


jrhender (changes requested):
@JGiter I believe there are several added files in the `setupUpIpfs` that should be removed.

.github/workflows/build-test.yml
```diff
@@ -46,7 +46,7 @@ jobs:
         uses: actions/cache@v2
         with:
           path: '**/node_modules'
-          key: ${{ runner.os }}-modules-${{ hashFiles('**/package.lock') }}
+          key: ${{ runner.os }}-modules-${{ hashFiles('./package-lock.json') }}
```
JGiter, line 49: `hashFiles` is failing now to hash all package-lock's. Probably its because we have too much of them. I think we can cashe only root package-lock, because it is enaugh to ensure same installation

nichonien (approved):

jrhender:
I think we should wait to merge this until we better understand the behaviour of the new IPFS Cluster Gateway's resolution of credentials on Infura gateway. Otherwise, resolution of credentials from ssi-hub might be too slow and could cause a large number of errors.

JGiter:
> @JGiter can you adjust this PR to be using the latest ew-did-registry code and target `ipfs-cluster` in tests is fine too) but instead change the `values.yaml` IPFS endpoint to be the Infura gateway? 

Yes, we can do this

devops/dev/values.yaml
```diff
@@ -159,4 +156,5 @@ iam-cache-server-helm:
...
-      IPFS_CLIENT_PROJECT_SECRET: AgCD3q5miwVO6GjALhdNMsbLQul35EQybREj9bV3IQJtC3ij7b65vpySapMAG1NBx6UMm8cqBRKySPxs85ICAevdtPXp9QtDLndTb/HiuEKC945Pw4R7g2V3RJaKgnKO/WZb2HRnaFq3ol63A4eIC2gcxAcU/M2Cinf3mQ8+Ha4D3ZwRTZBp+Vl7aD0vmDF8DrAaczBYs5qX0e4QYwCMlTviHGw/XNNKoQidl+jL+QGWE0aklOMrS9EBgcGtggYaBaNl5eqPAvhNllfsLIKAe4yt5QTQ2hCCih12/8HifOkAQAQw4l6xtzl3idfCSstwsMwMuAeBS/WUd40Y6tD1gR40AxSYiRC/FfMgDhW/ftqZD5wQq67YM7fD2BjJaM2UPDmzcyCdDqRGtHQNeCc9wf6iksGUv/4e68HRlrDQEw+DYqudeKcDs8jbQhJiecCDLGUr3Bb0Ww7Wi+tvs/rqtGrJzLyEJ5kM3XlEeuvA5qaGoJpEOa/0AnHLysfL3KnK+uA6t9KQTNUlRasmATIiHO0X0xSVJj0w30qgN6My83SbnIcbyno2rzlLktBxc+ZeB7pLb2ZvdFYYNgh05F0BXUMmqNXMAYerLpfZvAG4Y7JYcaPVkXnR16rG0vl0Qv+6/mVXbHV/Udfa8wjUrkbzl17mPGKZjmYdF8GoeGbjHsFM4JWceHsxbrX8d8m9NqPqYDRD420a6cqCFZUkpoq7RTa1hZUJ7HX5hlGyg75DuomDNQ==
+      IPFS_CLUSTER_ROOT: https://ewipfsgwtest.infura-ipfs.io/
```
jrhender, line 159: @JGiter I don't think this should be in the `sealedSecrets` section of values.yaml. Wouldn't it be better in the `appValues` section?

src/modules/claim/resolvers/revoker.resolver.ts
```diff
@@ -1,7 +1,9 @@
 import { IRevokerDefinition } from '@energyweb/credential-governance';
 import { RevokerResolver } from '@energyweb/vc-verification';
+import { Injectable } from '@nestjs/common';
 import { RoleService } from '../../role/role.service';
 
+@Injectable()
```
jrhender, line 6: @whitneypurdum I wonder if this might help fix the bug you've been looking into 👀

jrhender (approved):
Looks good, let's give it a try, thanks @JGiter

PR #517, opened 2022-10-14
task/SWTCH-1502 pin cached credentials

JGiter:
Added additional DidStore member in DidService, which configured to connect to IPFS cluster and responsible only for pinning. After migration will complete the first DidStore will be removed
Pinning of the claims is performed during resolution of the service endpoint, which happens dayily and on upding of document


nichonien (approved):

.env.ci
```diff
@@ -43,6 +43,7 @@ REDIS_PORT=$REDIS_PORT
 REDIS_HOST=$REDIS_HOST
 
 # IPFS
+IPFS_CLUSTER_ROOT=http://localhost:8080
```
jrhender, line 46: @JGiter why is `IPFS_CLUSTER_ROOT` added to `.env` files but `IPFS_CLUSTER_PASSWORD` is not?

JGiter: I didn't added because local IPFS isn't authorized

JGiter: Probably it is still worth adding as a hint

.env.ci
```diff
@@ -43,6 +43,8 @@ REDIS_PORT=$REDIS_PORT
 REDIS_HOST=$REDIS_HOST
 
 # IPFS
+IPFS_CLUSTER_ROOT=
+IPFS_CLUSTER_PASSWORD=
```
jrhender, line 47: @JGiter On some of the other entries, we have an $env var replacement. do we need that here too?
```suggestion
IPFS_CLUSTER_ROOT=$IPFS_CLUSTER_ROOT
IPFS_CLUSTER_PASSWORD=$IPFS_CLUSTERPASSWORD
```

JGiter: I thought that to test some environment locally one will set secret values directly in .env.* Do you think .env better to populate from other source?

src/modules/did/did.service.ts
```diff
@@ -430,13 +442,20 @@ export class DIDService implements OnModuleInit, OnModuleDestroy {
...
+        ) {
+          return cachedService;
+        }
 
         if (!IPFSService.isCID(serviceEndpoint)) {
           return { serviceEndpoint, ...rest };
         }
 
         const token = await this.didStore.get(serviceEndpoint);
+        await this.didCluster.save(token);
```
jrhender, line 458: @JGiter Could you explain a little more your rationale for this approach?

I think this definitely works well, thanks for your work thus far 🙇 
However, I think we should consider if it is worth improving this a bit. My thinking is that:
- 👍  This approach ensure that every thing that `resolveNotCachedClaims` is called, that a credential which is not pinned in the new cluster will be pinned
- 👎  It unnecessarily slows down DID document resolution because there is extra pinning step, which the client doesn't really care about.
- 👎  It adds a side-effect to `resolveNotCachedClaims`. Related to this, IMO, the approach causes `resolveNotCachedClaims` violates the single responsibility principle.
-  👎 It threads the migration code through `resolveNotCachedClaims` such that it might be difficult for someone not familiar with the code to know which part to surgically remove to stop migration (they would probably need to go back and look at this PR)

What do you think of an alternative approach of creating a new private method, maybe called `pinAllServices`, that can accept an array of `IClaim` and pin to the cluster. I know this would mean potentially retrieving the credential twice from IPFS, but I think it would make the code a bit clearer. What are your thoughts?

JGiter: I agree that `resolveClaims` isn't proper place for pinning. I think pinning is more belong to caching. If so are `incrementalRefreshCachedDocument` and `addCachedDocument` more suitable for pinning?

> It unnecessarily slows down DID document resolution

Can we pin in job queue?

JGiter: > What do you think of an alternative approach of creating a new private method, maybe called pinAllServices

The queue is preferable, because it doesn't add delay to document resolution

src/modules/did/did.processor.ts
```diff
@@ -23,16 +39,30 @@ export class DIDProcessor {
...
-      await this.didService.incrementalRefreshCachedDocument(job.data);
+      doc = await this.didService.incrementalRefreshCachedDocument(job.data);
+    }
+
+    await this.pinClaims(doc.service.map((s) => s.serviceEndpoint));
+  }
+
+  private async pinClaims(cids: string[]) {
+    for await (const cid of cids) {
+      if (!this.didCluster.isPinned(cid)) {
```
JGiter, line 62: Didn't use here Promise.all, because it would make code more havyweight: three Promise.all for `isPinned`, 'get', `save` and `isPinned` along with pin status need to pass cid

src/modules/did/did.service.ts
jrhender: I think you're right, I think that `processDIDDocumentAddition` and `processDIDDocumentRefresh` will only be called from the `ADD_DID_DOC_QUEUE_NAME` and the `UPDATE_DID_DOC_QUEUE_NAME` queues. So I think we can leave it as is.

src/modules/did/did.processor.ts
jrhender: Ok, sounds good

src/modules/did/did.processor.ts
```diff
@@ -23,16 +39,30 @@ export class DIDProcessor {
...
+      doc = await this.didService.addCachedDocument(job.data, true);
     } else {
-      await this.didService.incrementalRefreshCachedDocument(job.data);
+      doc = await this.didService.incrementalRefreshCachedDocument(job.data);
+    }
+
+    await this.pinClaims(doc.service.map((s) => s.serviceEndpoint));
+  }
+
+  private async pinClaims(cids: string[]) {
```
jrhender, line 60: What do you think of adding this documentation
```suggestion
  /**
   * This method migrates claims by retrieving from one DidStore and pinning to another
   * It was implemented for EW migration from Infura to EW hosted IPFS
   */
  private async pinClaims(cids: string[]) {
```

src/modules/did/did.processor.ts
```diff
@@ -23,16 +39,30 @@ export class DIDProcessor {
...
+    for await (const cid of cids) {
```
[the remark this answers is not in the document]
JGiter: I think there is important difference here. Unlike regular `for` the `await` variance awaits result of each iteration. This is stated in the `Description` section

> If the iterable is a sync iterable that yields promises, for await...of would produce a sequence of resolved values, while for...of would produce a sequence of promises

This is important for us, because we expect all claims are persisted when job complets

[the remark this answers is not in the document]
jrhender: Hmm, I think I'm still missing your point. The quote you mentioned is about the `iterable`, referring to this code snippet from the documentation https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for-await...of#syntax
```
for await (variable of iterable)
  statement
```

And the `iterable` in our case is `cids` which is of type `string[]`, so the `iterable` is not a "sync iterable that yields promises", right?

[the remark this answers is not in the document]
JGiter: Yes, you are right. I confused promise yieled by iterator with loop body

nichonien (approved):

devops/prod/values.yaml
```diff
@@ -171,3 +171,6 @@ iam-cache-server-helm:
...
+      IPFS_CLUSTER_PASSWORD: AgBHwmoO6LJ3hCAYMFxjeZyDdsAERdupnUHZs87nV3PRjIQ2ZQRCZk+6MrpgoReNpyu6cteefF3rGrgW2GHqSmgvfKOGzIPwQRCj5XLaW/dKf1dtwFKFzyFBGOfeZ2QkKzgMXz2APdulPoUAmAlri7p/B7lp5CjbzJ3+Tyildc/JSFKwIL8gtRUEPdtIqzokmWp7HQ1P2NJfM6tas2KxC8JqrKDC8pqutft4C5RN1p8Grsyi2kW/oDUhwLDgl8oJJnhu7GvgelAugpyvWFzCw6KBXgYs5cydRsAZ1caQZbrgtZcKCi0mxbu0omqr4o0bMNuinUgpa9Ib7aBsH+dRHgAbyYPv/HL0jk9yuoA2mnKQ+bu8VaLZcwxRTfTqyreTlYYukEnYDTspXsj9DcdRf/hyfYLtYegaCnSID/cY+nSOLrqc0KdtxeGiJuc6msPYQI1qOjXVHPwFJSCEjb2Aph10lSNumzsUC9kMfXePi0lb1LRkdOuKQojI5oQfij06V+eHyaaJpsiRgh9ZNtOfZEoniYWPMfh/NwS/FewRvTD6l6Prr1ZVbv/d1mLoy+T0WAsRSPn5fpiUIGNvErRNEYM5g3WLlEmH/Zca+7eiFL+suR03pJdpylKF4JQ+MD15DKbGNtdEQHfJHG+AKlMRexMItkvWFTdL/bYgo4Xhf2bAFv+yhZ+T27ld8vgX8xg/VXRDbdzC62+kZS4cvW7GnMz6qDdbRg==
```
jrhender, line 176: @JGiter I don't think these can be the right values, as we don't even have the prod cluster up yet, right?

JGiter: right. I thought we will have one ipfs for all environments like when using Infura. Should I leave IPFS cluster config only in dev env?

jrhender: > Should I leave IPFS cluster config only in dev env?

Yes, I think so, as we don't have it set up for other environments

jrhender (approved):
Looks good, let's try it 👍

nichonien (approved):

PR #520, opened 2022-10-14
Make RoleRevokerResolver injectable

whitneypurdum:


whitneypurdum:
In attempts to address this bug (maybe the revoker needs to be injected rather than instantiated)  https://energyweb.atlassian.net/jira/software/projects/SWTCH/boards/54?assignee=608057a2e3b59800688dbc2a&selectedIssue=SWTCH-1539

whitneypurdum:
@jrhender requesting your review as Friday you suggested this may be the bug for revoking by role.

JGiter (approved):

src/modules/claim/services/revocation-verification.service.ts
```diff
@@ -15,10 +15,10 @@ export class RevocationVerificationService extends RevocationVerification {
     didService: DIDService,
     roleService: RoleService,
     provider: Provider,
-    @Inject('RegistrySettings') registrySettings: RegistrySettings
+    @Inject('RegistrySettings') registrySettings: RegistrySettings,
+    revokerResolver: RoleRevokerResolver
   ) {
     const issuerResolver = new RoleIssuerResolver(roleService);
```
jrhender, line 21: Hopefully this worked 🤞 . But looking at it now, it seems a bit odd to only inject `RoleRevokerResolver` and not `RoleIssuerResolver`. What are your thoughts on this @whitneypurdum @JGiter ?

PR #525, opened 2022-10-21
#swtch 1583: enrolment precondition credentials are not being verified

whitneypurdum:
For task https://energyweb.atlassian.net/jira/software/projects/SWTCH/boards/54?assignee=608057a2e3b59800688dbc2a&selectedIssue=SWTCH-1583

This PR:
1. adds a method in claim service's  handleClaimEnrolmentRequest to verify claims that are enrolment pre-requisites for a role the user is actively requesting enrolment to
2. Creates a claim-verification service that resolves a credential (EIP191) and verifies proof, expiration, issuer verification (mirrors iam client lib process)
3. Adds tests for  handleClaimEnrolmentRequest and tests for claim-verification service
4. Adds IssuerVerificationService that is necessary for the Claim Verification
5. moves the method for verifying that a claim is present in a user's did document (verifyClaimPresentInDidDocument) out of role service as per @JGiter review suggestion


src/modules/claim/services/claim-verification.service.ts
```diff
@@ -0,0 +1,123 @@
...
+   */
+  public async resolveCredentialAndVerify(
+    subjectDID: string,
+    roleNamespace: string
+  ) {
+    const resolvedCredential = await this.credentialResolver.getCredential(
+      subjectDID,
+      roleNamespace
+    );
+    if (!resolvedCredential) {
```
whitneypurdum, line 40: Hi @nichonien quick q for you - I am pretty sure looking at your code that it will always return an EIP first for a resolved credential, but want to make sure - if not I will need to add logic here to also verify a JSON LD credential too, but I think priority is always given to JWT, right? Do you think this is sound logic?

nichonien: It is designed to return both a `VC` and `RoleEIP191Jwt`, the outcome changes based on how the `CredentialResolver` is configured. 

But on a major note, it is mostly relying on DIDDocument for resolution (be it from cached DID Document or directly from blockchain) and thus for now it is just `RoleEIP191Jwt` (as only this type of credential are added to DID Document).

nichonien: If you look at how it is done in `passport-did-auth`  https://github.com/energywebfoundation/passport-did-auth/blob/develop/lib/RoleCredentialResolver.ts

This can return both VC and EIP credentials, but because only EIP credentials are being published to DID Document it will always return EIP type credential (this might / would change in future).

whitneypurdum: Thanks @nichonien. So this PR should also add the method to verify a VC as well, correct? Not just a jwt?

nichonien: Yeah, we can add but I am not sure if it is a must as we are issuing both VC and jwt credentials for a role. This way we will always be sure that verifying only RoleEIP191 credentials will validate all scenarios. @jrhender what are your thoughts ?

jrhender: IMO, just EIP191JWT is enough

whitneypurdum: Great, thanks both!

whitneypurdum: @JGiter do you know the trick to getting tests run locally to behave like tests run in the pipeline? I am running npm run test locally and all my tests are passing, but are failing in pipeline. Just want to recreate locally so I can test it locally.

src/modules/claim/services/claim-verification.service.ts
```diff
@@ -0,0 +1,123 @@
...
+import { DIDService } from '../../did/did.service';
+import { RoleCredentialResolver } from '../resolvers/credential.resolver';
+import { ProofVerifier } from '@ew-did-registry/claims';
+import {
+  CredentialResolver,
+  RoleEIP191JWT,
+  isEIP191Jwt,
+} from '@energyweb/vc-verification';
+
+@Injectable()
```
whitneypurdum, line 13: @jrhender or @JGiter could you please let me know if my approach to this issue was okay. Not sure if creating/injecting another service (which requires issuer verification service creation as well) is okay.

src/modules/claim/services/claim-verification.service.ts
JGiter: May be it is because DID_REGISTRY_ADDRESS isn't set in CI https://github.com/energywebfoundation/ssi-hub/blob/73d6553a820fb9c7b20ed730118a4d8bc9cf41ce/.github/workflows/build-test.yml#L69, but set in your local .env

src/modules/claim/services/claim-issuance.service.ts
```diff
@@ -83,10 +85,41 @@ export class ClaimIssuanceService {
       claimType: dto.claimType,
     });
 
-    await this.roleService.verifyEnrolmentPrecondition({
-      claimType: dto.claimType,
-      userDID: dto.requester,
-    });
+    const enrolmentPreconditions =
```
JGiter, line 88: It would be easer to read and test if verification of enrollment precoditions was moved in separate method

src/modules/role/role.service.ts
```diff
@@ -205,39 +205,50 @@ export class RoleService {
...
-          );
-        }
-      }
+  /**
+   * Verifies that a user's Did Document contains all roles required for enrolment (enrolment preconditions)
+   * @param claimType the role to verify enrollment preconditions for
+   * @param userDID the Did of the user seeking to obtain the role. Must posses enrolment preconditions
+   * @param conditions enrolment preconditions needed to obtain the role (claimType)
+   */
+  public async verifyDidDocumentContainsEnrolmentPreconditions({
```
JGiter, line 235: From my perspective verification belongs to `claim` module

src/modules/claim/services/claim-verification.service.ts
whitneypurdum: Thanks @JGiter, do you mean I should change the filename to claims.controller.spec.ts from claims.controller.e2e.spec.ts

JGiter: Yes

whitneypurdum: @JGiter there is [already a claim.controller.spec](https://github.com/energywebfoundation/ssi-hub/blob/develop/src/modules/claim/claim.controller.spec.ts) (which really doesn't test anything). Should I just move everything from claims.controller.e2e.spec.ts to this test file?

whitneypurdum: Removed e2e and now just claim.controller.spec

src/modules/role/role.service.ts
whitneypurdum: got it. I moved this to the claim-verification service. Lmk if you think it is not good there.

src/modules/claim/claim.service.spec.ts
```diff
@@ -1,15 +1,144 @@
...
+const provider = new MockProvider();
+let didRegistry: EthereumDIDRegistry;
+const cachedIdentity = provider.getWallets()[0];
 
+const MockConfigService = {
+  get: jest.fn((key: string) => {
+    if (key === 'DID_SYNC_ENABLED') {
+      return 'false';
+    } else if (key === 'DID_REGISTRY_ADDRESS') {
+      return didRegistry.address;
```
whitneypurdum, line 59: @JGiter I was planning on adding tests to the claim.service test file, if this sounds okay with you.

src/modules/claim/claim.controller.ts
```diff
@@ -178,6 +178,7 @@ export class ClaimController {
     const claimDTO = ClaimRequestDTO.create(claimData);
 
     await validateOrReject(claimDTO);
+    //THIS IS WHERE IT IS!!
     const result = await this.claimService.handleClaimEnrolmentRequest(
```
whitneypurdum, line 182: adding test in claim.service tests for this method handleClaimEnrolmentRequest

src/modules/claim/services/claim-verification.service.ts
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  /**
+   * Resolve a credential from storage and verify its proof/signature and its issuer's authority
+   *
+   * @param subjectDID The DID to try to resolve a credential for
+   * @param roleNamesapce The role to try to get a credential for. Should be a full role namespace (for example, "myrole.roles.myorg.auth.ewc")
+   * @return void. Returns boolean indicating if credential is verified. Contains array of error messages if not verified.
+   */
+  public async resolveCredentialAndVerify(
```
whitneypurdum, line 32: @jrhender Should I be adding a test suite for this file for each of these methods?

whitneypurdum: I went ahead and did tests for these methods.

src/modules/claim/claim.controller.spec.ts
```diff
@@ -1,19 +1,589 @@
-import { Test, TestingModule } from '@nestjs/testing';
+import { ExecutionContext, INestApplication } from '@nestjs/common';
```
whitneypurdum, line 1: just changed the name of this file to not be e2e as Dmitry and I discussed

whitneypurdum:
This PR:
1. adds a method in claim service's  handleClaimEnrolmentRequest to verify claims that are enrolment pre-requisites for a role the user is actively requesting enrolment to
2. Creates a claim-verification service that resolves a credential (EIP191) and verifies proof, expiration, issuer verification (mirrors iam client lib process)
3. Adds tests for  handleClaimEnrolmentRequest and tests for claim-verification service

src/modules/claim/services/claim-issuance.service.ts
```diff
@@ -83,16 +86,70 @@ export class ClaimIssuanceService {
...
+      await this.roleService.fetchEnrolmentPreconditions({
+        claimType: dto.claimType,
+      });
+    if (
+      enrolmentPreconditions?.length > 0 &&
+      enrolmentPreconditions.find((precond) => precond.type === 'role')
```
jrhender, line 95: This line is interesting. 
- At a minimum, it should be replaced with the enum rather than the string
- IMO, we need to add an else statement that throws error if the precond type is not 'role'. Otherwise, if another type is added and someone runs ssi-hub that doesn't recognize it, enrolmentPrecondition check could be bypassed. IMO, it is better to err on the side of safety here.
- Given the above point, I think we should use `every` and not `find`.
```suggestion
      enrolmentPreconditions.every((precond) => precond.type === PreconditionType.Role)
```

src/modules/claim/services/claim-issuance.service.ts
```diff
@@ -83,16 +86,70 @@ export class ClaimIssuanceService {
...
+  public async verifyEnrolmentPrerequisites(
+    enrolmentPreconditions: {
+      type: PreconditionType;
+      conditions: string[];
+    }[],
+    requester: string,
+    claimType: string
+  ) {
+    for (const { type, conditions } of enrolmentPreconditions) {
+      if (type === 'role' && conditions?.length > 0) {
```
jrhender, line 117: I think this could be seen as a duplication of the check above (the `enrolmentPreconditions.find`). I think probably only one of them are needed

jrhender (changes requested):
It's a solid start in the right direction but a bit more discussion is needed I think.

src/modules/claim/services/claim-verification.service.ts
```diff
@@ -0,0 +1,156 @@
...
+      eip191Jwt,
+      payload?.iss as string
+    );
+    if (!proofVerified) {
+      errors.push(
+        `Verification failed for ${roleNamespace} for ${subjectDID}: Proof not verified for role`
+      );
+    }
+    // Date.now() and JWT expiration time both identify the time elapsed since January 1, 1970 00:00:00 UTC
+    const isExpired = payload?.exp && payload?.exp < Date.now();
```
whitneypurdum, line 96: @whitneypurdum  DO NOT MERGE BEFORE VERIFYING THIS!! (*1000)

src/modules/claim/services/claim-issuance.service.ts
whitneypurdum: used enum, made .every, made error message to throw if .every was not met and enrolmentConditions.length

src/modules/role/role.service.ts
```diff
@@ -205,40 +205,25 @@ export class RoleService {
...
 
     const {
       definition: { enrolmentPreconditions },
     } = role;
-
-    if (!enrolmentPreconditions || enrolmentPreconditions.length < 1) return;
-    for (const { type, conditions } of enrolmentPreconditions) {
-      if (type === 'role' && conditions && conditions?.length > 0) {
-        const conditionMet = didDocument.service.some(
-          ({ claimType }) =>
```
whitneypurdum, line 232: @jrhender or @JGiter is this really a good check if the condition is met? I feel like it should map through each condition and find it in the service array, not use .some on service array, right?

src/modules/claim/services/claim-verification.service.ts
```diff
@@ -0,0 +1,193 @@
...
+    userDID: string;
+    conditions: string[];
+  }) {
+    const didDocument = await this.didService.getById(userDID);
+    return didDocument.service.some(
+      ({ claimType }) => claimType && conditions.includes(claimType as string)
+    );
+  }
+
+  public async verifyEnrolmentPreconditions(
```
whitneypurdum, line 142: I moved the duplicative logic for checking conditions out of claim and claim-issuance service into this file, so they can share the same logic and not repeat.

src/modules/role/role.service.ts
whitneypurdum: (Asking because this logic was moved to claim-verification service, above.)

JGiter: @whitneypurdum It depends on `mustHaveAll` https://github.com/energywebfoundation/ew-credentials/blob/861feac6593b4807f1fc72e9f184357f7ac007fd/packages/credential-governance/contracts/profiles/enrolment-conditions/EnrolmentPrerequisiteRolesResolver.sol#L13

whitneypurdum: @JGiter I still don't quite get it. This task is to ensure that the user 1. has the pre-requisite roles needed to take on a new role and 2. that all of those roles are valid. Should we change this so that this makes sure that every pre-requisite role that is needed is in the DID document?

JGiter: @whitneypurdum the enrollment preconditions are set in role definition. Depending on `mustHaveAll` preconditions are met when either one or all conditions are met. So the `.some` on service array should be used when `mustHaveAll` is `false` and `.every` should be otherwise

whitneypurdum: OKay, thanks, I will leave as is then.

src/modules/claim/services/claim-verification.service.ts
```diff
@@ -0,0 +1,181 @@
...
+   * Resolve a credential from storage and verify its proof/signature and its issuer's authority
+   *
+   * @param subjectDID The DID to try to resolve a credential for
+   * @param roleNamesapce The role to try to get a credential for. Should be a full role namespace (for example, "myrole.roles.myorg.auth.ewc")
+   * @return void. Returns boolean indicating if credential is verified. Contains array of error messages if not verified.
+   */
+  public async resolveCredentialAndVerify(
+    subjectDID: string,
+    roleNamespace: string
+  ) {
```
jrhender, line 32: I think that it would be best to have a return type on the function (I'm not sure I got the syntax right)
```suggestion
  ): Promise<{ isVerified: boolean, errors: string[] }> {
```

src/modules/claim/services/claim-verification.service.ts
```diff
@@ -0,0 +1,181 @@
...
+      return {
+        isVerified: false,
+        errors: [
+          `No credential found for role ${roleNamespace} for Did ${subjectDID}`,
+        ],
+      };
+    }
+    if (isEIP191Jwt(resolvedCredential)) {
+      return this.verifyRoleEIP191JWT(resolvedCredential as RoleEIP191JWT);
+    }
```
jrhender, line 47: I think we should have a more expected return here, rather than falling through to `undefined` 
```suggestion
    if (isEIP191Jwt(resolvedCredential)) {
      return this.verifyRoleEIP191JWT(resolvedCredential as RoleEIP191JWT);
    }
    return {
      isVerified: false,
      errors: [
        `credential found for role ${roleNamespace} but unable to verify for Did ${subjectDID}`,
      ],
    }
```

src/modules/claim/resolvers/credential.resolver.ts
```diff
@@ -74,9 +74,13 @@ export class RoleCredentialResolver implements CredentialResolver {
    */
   serviceEndpointsToEIP191(tokens: string[]): RoleEIP191JWT[] {
     return tokens
-      .map((token) => {
+      ?.map((token) => {
```
jrhender, line 77: I'm curious on your rationale for the optional chaining operator here. I think I might prefer to have an exception be thrown here if `tokens` was `undefined` rather than `undefined` be returned. What are your thoughts here?

src/modules/claim/claim-verification.service.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+          'claimType'
+        );
+        expect(result).toThrowError(
+          `Role enrolment precondition not met for user: ${
+            resolvedCredential.payload.did
+          } for role: ${'conditionone.org.ewc'}. Verification errors for enrolment preconditions: ${JSON.stringify(
+            mockError.errors
+          )}`
+        );
+      } catch (_) {}
```
jrhender, line 145: why is the test in this a `try/catch` block? I thought that `toThrowError` would handle the thrown error. I'm worried that a try/catch block could swallow unexpected errors and cause the test to pass despite this error.

src/modules/claim/services/claim-verification.service.ts
```diff
@@ -0,0 +1,181 @@
...
+        payload?.claimData?.claimType
+      );
+    if (!issuerVerified && error) {
+      throw new Error(`Issuer verification failed: ${error}`);
+    }
+    return {
+      errors: errors,
+      isVerified: !!proofVerified && issuerVerified && !isExpired,
+    };
+  }
```
jrhender, line 94: Is revocation being considered here? If not, I think it would be good to add, though maybe we could do it in a separate PR.

jrhender (commented):
Some further comments

src/modules/claim/services/claim-verification.service.ts
whitneypurdum: @jrhender This will return the errors from verifyRoleEIP191JWT, which has the isVerifiedMessage and the errors, and the reason. return this.verifyRoleEIP191JWT(resolvedCredential as RoleEIP191JWT) returns
 return {
      errors: errors,
      isVerified: !!proofVerified && issuerVerified && !isExpired,
    };

src/modules/claim/resolvers/credential.resolver.ts
whitneypurdum: removed

src/modules/claim/services/claim-verification.service.ts
whitneypurdum: added return type

src/modules/claim/services/claim-verification.service.ts
whitneypurdum: @jrhender revocation check is performed in the verifyIssuer method in ew-credentials.

src/modules/claim/services/claim-verification.service.ts
```diff
@@ -0,0 +1,187 @@
...
+
+  /**
+   * Verifies:
+   * - That off-chain claim was issued by authorized issuer
+   * - That claim is not expired
+   * - That off-chain claim proof is valid
+   *
+   * @param roleEIP191JWT off chain claim to verify
+   * @param subjectDID The credential subject
+   * @param roleNamesapce The role to try to get a credential for. Should be a full role namespace (for example, "myrole.roles.myorg.auth.ewc")
```
nichonien, line 58:
```suggestion
```

src/modules/claim/services/claim-verification.service.ts
jrhender: Sorry, I don't quite get your point. `this.verifyRoleEIP191EWT` is only called if `isEIP191Jwt` is true. If `isEIP191Jwt` is `false`, then this will just return `undefined`, I believe. It's this return of undefined that I think is best to avoid so that the return object is more consistent. Otherwise, I think we need to change the return type of the method to `Promise<{ isVerified: boolean; errors: string[] } | undefined>`.

whitneypurdum: OKay, should the error then indicate that an unsupported credential type was resolved (i.e. not EIP191?) (indicating that a credential was found but we don't support verification methods for it?) @jrhender

src/modules/claim/services/claim-verification.service.ts
whitneypurdum: @nichonien good catch, thank you!

src/modules/claim/services/claim-verification.service.ts
whitneypurdum: Just thinking that 'unable to verify for {did}' would make me think theres an issue with my DID rather than the credential itself, but maybe I'm overthinking it.

nichonien: I think something like `verification not supported for present credential type` or `No EIP191 credential found for the given namespace` should be more clear.

jrhender: @whitneypurdum yes, I think either of @nichonien's suggestions would be good

whitneypurdum: @jrhender This has been udpated.

src/modules/claim/claim.controller.spec.ts
```diff
@@ -1,19 +1,589 @@
...
-  let controller: ClaimController;
+const redisConfig = {
+  port: parseInt(process.env.REDIS_PORT),
+  host: process.env.REDIS_HOST,
+  password: process.env.REDIS_PASSWORD,
+};
 
-  beforeEach(async () => {
-    const module: TestingModule = await Test.createTestingModule({
+describe('ClaimsController', () => {
```
jrhender, line 57: Thanks for getting unit tests going for the `ClaimsController`!

src/modules/claim/services/claim-verification.service.ts
jrhender: Thanks, looks good

src/modules/claim/services/claim-verification.service.ts
```diff
@@ -0,0 +1,192 @@
...
+    const verifier = new ProofVerifier(didDoc);
+    return verifier.verifyAssertionProof(token);
+  }
+
+  /**
+   * Verifies that a user's Did Document contains all roles required for enrolment (enrolment preconditions)
+   * @param userDID the Did of the user seeking to obtain the role. Must posses enrolment preconditions
+   * @param conditions enrolment preconditions needed to obtain the role (claimType)
+   */
+  public async verifyClaimPresentInDidDocument({
```
jrhender, line 122: Minor point, but I personally like having the public method of a class before the private methods. I proposed it as a code convention: https://energyweb.atlassian.net/wiki/spaces/MYEN/pages/3207626825/Code+Conventions+and+Guidelines

jrhender (approved):
This looks good to me, thanks @whitneypurdum

nichonien (approved):

src/modules/claim/services/claim-verification.service.ts
whitneypurdum: @jrhender okay, I moved it to top above private mthds.

jrhender: Thanks. The intuition as that as someone reads a Class file, they are first interested in the public interface of the class and then they can continue reading if they want to know more implementation details

PR #528, opened 2022-10-24
#swtch 1539 revoke by role

whitneypurdum:
Fix for bug https://energyweb.atlassian.net/jira/software/projects/SWTCH/boards/54?assignee=608057a2e3b59800688dbc2a&selectedIssue=SWTCH-1539

Fixes 2 issues:
1. Uses arrow in chained methods in order to preserve "this" context (fixes issue of cannot read 'ipfs' of undefined or this)
2. Formats token so it can be passed to isEIP191Jwt method. Previously was not formatted and so no claims were coming back from the check, and therefore the pre-requisite role was never found for verifier

Regression testing performed:
1. Can revoke on-chain credential for role-based revoker
2. Can revoke off-chain credential for did-based revoker
3. Can revoke on-chain credential for did-based revoker

**I'm thinking we won't want to merge this till after demo tomorrow, just in case, even if approved**

### DIFF
--- a/src/modules/claim/resolvers/credential.resolver.ts
+++ b/src/modules/claim/resolvers/credential.resolver.ts
@@ -76,7 +76,11 @@ export class RoleCredentialResolver implements CredentialResolver {
     return tokens
       .map((token) => {
         try {
-          return jwt.decode(token) as RolePayload;
+          const decoded = jwt.decode(token) as RolePayload;
+          return {
+            eip191Jwt: token,
+            payload: decoded,
+          };
         } catch (_) {
           return {};
         }
--- a/src/modules/did/did.service.ts
+++ b/src/modules/did/did.service.ts
@@ -331,8 +331,12 @@ export class DIDService implements OnModuleInit, OnModuleDestroy {
     return Promise.all(
       service
         .map(({ serviceEndpoint }) => serviceEndpoint)
-        .filter(IPFSService.isCID)
-        .map(this.didStore.get)
+        .filter((endpoint) => {
+          return IPFSService.isCID(endpoint);
+        })
+        .map((did) => {
+          return this.didStore.get(did);
+        })
     );
   }
 
--- a/src/modules/did/did.service.ts
+++ b/src/modules/did/did.service.ts
@@ -331,12 +331,8 @@ export class DIDService implements OnModuleInit, OnModuleDestroy {
     return Promise.all(
       service
         .map(({ serviceEndpoint }) => serviceEndpoint)
-        .filter((endpoint) => {
-          return IPFSService.isCID(endpoint);
-        })
-        .map((did) => {
-          return this.didStore.get(did);
-        })
+        .filter((endpoint) => IPFSService.isCID(endpoint))
+        .map((cid) => this.didStore.get(cid))
     );
   }
 


src/modules/claim/resolvers/credential.resolver.ts
```diff
@@ -76,7 +76,11 @@ export class RoleCredentialResolver implements CredentialResolver {
     return tokens
       .map((token) => {
         try {
-          return jwt.decode(token) as RolePayload;
+          const decoded = jwt.decode(token) as RolePayload;
+          return {
+            eip191Jwt: token,
```
JGiter, line 81: Are you sure that `token` should be RolePayload? I thought we are going to store VC/EIP191 on IPFS. If that is right then error is casting decoded token to `RolePayload`

jrhender: @JGiter further down in this method (on line 88), there is a `.filter(isEIP191Jwt)` which is looking for both the `eip191Jwt` property and `payload.claimData.claimType`. https://github.com/energywebfoundation/ew-credentials/blob/861feac6593b4807f1fc72e9f184357f7ac007fd/packages/vc-verification/src/utils/formatters.ts#L98
So it makes sense to me that this would fix the issue.

src/modules/did/did.service.ts
```diff
@@ -331,8 +331,8 @@ export class DIDService implements OnModuleInit, OnModuleDestroy {
     return Promise.all(
       service
         .map(({ serviceEndpoint }) => serviceEndpoint)
-        .filter(IPFSService.isCID)
-        .map(this.didStore.get)
+        .filter((endpoint) => IPFSService.isCID(endpoint))
+        .map((cid) => this.didStore.get(cid))
```
jrhender, line 335: Great thinking @JGiter and nice perseverance @whitneypurdum . We will definitely be better equipped to realized JS `this` context issues in the future 👍

src/modules/claim/resolvers/credential.resolver.ts
whitneypurdum: Hmm I think so. This is the return from the decode and below that the RolePayload interface
{
 claimData: {
  requestorFields: [],
  claimType: ‘errortestaug1.roles.suborgs.whitney.iam.ewc’,
  claimTypeVersion: 1,
  issuerFields: []
 },
 did: ‘did:ethr:volta:0x17b65C8C9746F87c82cc6f7C4FC38fCA5f1AeB75’,
 signer: ‘did:ethr:volta:0x17b65C8C9746F87c82cc6f7C4FC38fCA5f1AeB75’,
 credentialStatus: {
  id: ‘https://identitycache-dev.energyweb.org/v1/status-list/urn:uuid:5829660e-46b1-4ee5-8072-0db6e67ef7e7’,
  type: ‘StatusList2021Entry’,
  statusPurpose: ‘revocation’,
  statusListIndex: ‘0’,
  statusListCredential: ‘https://identitycache-dev.energyweb.org/v1/status-list/urn:uuid:5829660e-46b1-4ee5-8072-0db6e67ef7e7’
 },
 iss: ‘did:ethr:volta:0x17b65C8C9746F87c82cc6f7C4FC38fCA5f1AeB75’,
 sub: ‘did:ethr:volta:0x17b65C8C9746F87c82cc6f7C4FC38fCA5f1AeB75’,
 exp: 1659462891
}

export interface RolePayload extends JwtPayload {
    credentialStatus?: StatusList2021Entry;
    claimData: ClaimData;
    signer: string;
}

And then that object is used here:
```
 this.serviceEndpointsToEIP191(resolvedEndpoints).find(
        (token) => token.payload.claimData.claimType === namespace
      )
```
Whereas before it was just sending the decoded token, so it didn't have the key token on the object, so it wasn't finding a match.

jrhender (approved):
Excellent PR description, thanks @whitneypurdum 👍 

Re
> I'm thinking we won't want to merge this till after demo tomorrow, just in case, even if approved

Yes, that's the case. I've added the corresponding label.

Another question I have is what we should do with this previous fix attempt: https://github.com/energywebfoundation/ssi-hub/pull/520#discussion_r997927510
Do you think it would make sense to revert it? Or better to leave it as is?

whitneypurdum:
> Excellent PR description, thanks @whitneypurdum 👍
> 
> Re
> 
> > I'm thinking we won't want to merge this till after demo tomorrow, just in case, even if approved
> 
> Yes, that's the case. I've added the corresponding label.
> 
> Another question I have is what we should do with this previous fix attempt: [#520 (comment)](https://github.com/energywebfoundation/ssi-hub/pull/520#discussion_r997927510) Do you think it would make sense to revert it? Or better to leave it as is?

Hmm good q - the idea from those injections were taken from a PR that Dmitry had open, but I think reverted. @JGiter Should I undo the injections from these classes you think? If I'm not mistaken it will be necessary when we bump ew-credentials version later on, right? @JGiter thanks

PR #529, opened 2022-10-31
feat: developer experience improvements

artursudnik:
This PR implements a number of changes to make it easier start the `si-hub` locally by a new developer


.env.ci
```diff
@@ -10,24 +10,29 @@ DB_NAME=$DB_NAME
 
 TYPEORM_LOGGING=error,migration,warn,info
 
+# REDIS
+REDIS_VERSION=6.0.8
+REDIS_PASSWORD=$REDIS_PASSWORD
+REDIS_PORT=$REDIS_PORT
+REDIS_HOST=$REDIS_HOST
```
JGiter, line 17: Shouldn't Redis be launched on localhost, when devloping locally?

docs/api/README.md
```diff
@@ -11,15 +11,16 @@ This repository was formerly refered to as the `iam-cache-server`.
 
 ## Prepare
 
+### Development
+
 ```bash
-# development
 $ git checkout develop
 $ cp .env.dev .env
```
JGiter, line 18: We also need to mention to populate IPFS configuration

.env.dev
```diff
@@ -32,13 +32,17 @@ STAKING_POOL_FACTORY_ADDRESS=0x4b2A127680320eD980beAa7aD9b2447B96BC32fC
...
 ASSETS_SYNC_ENABLED=true
 
 # IPFS
+IPFS_CLIENT_HOST=
+IPFS_CLIENT_PORT=
+IPFS_CLIENT_PROJECT_ID=
+IPFS_CLIENT_PROJECT_SECRET=
 
 # INTERVALS
 DIDDOC_SYNC_INTERVAL_IN_HOURS=1
```
JGiter, line 48: Do you think that local ssi-hub should be synchronized with dev? If so, then we should probably trigger synchronization immediately after ssi-hub is launched. 
May be it is enaugh to cache only DID used for login

.env.dist
```diff
@@ -22,14 +22,14 @@ NATS_ENVIRONMENT_NAME=
 # ENS
 CHAIN_ID=73799
 CHAIN_NAME=volta
-ENS_URL=https://volta-rpc-vkn5r5zx4ke71f9hcu0c.energyweb.org/
+ENS_URL=https://volta-rpc-vkn5r5zx4ke71f9hcu0c.energyweb.org
```
JGiter, line 25: I beleive production environment corresponds to EWC. So here should be private EWC URL

JGiter (commented):
From my opinion it would be also useful to provide config for `production`
We should also remove private RPC's

.env.ci
artursudnik: it is: https://github.com/energywebfoundation/ssi-hub/blob/543ed0cfffdfd882ca6a790996b6c3e3b1748a64/docker-compose.dev.yml#L28-L28

docs/api/README.md
artursudnik: Good point.

.env.dev
artursudnik: I am not sure. @jrhender , what is your opinion?

nichonien: Yeah, IMO caching only the DID used for login / testing would be enough.

.env.ci
JGiter: `.env.ci` and `compose.dev.yml` references `REDIS_*` variables, but where these variables are actually set?

artursudnik: I believe that `$REDIS_HOST` and `$REDIS_PORT` are set in GH action. I just moved this part of this file to make its structure the same as .env.dev. It is easier to compare them now.

.env.dev has just values here: https://github.com/energywebfoundation/ssi-hub/blob/543ed0cfffdfd882ca6a790996b6c3e3b1748a64/.env.dev#L12-L12

JGiter: we don't use .env in GH now. These files serves only as config for local develop/testing

artursudnik: So, probably `.env.ci` is not necessary at all and could be removed to not distract people. What do you think, @JGiter

JGiter: Yes, we don't have `ci` environment. We would better replace .env.ci with .env.staging, .env.dist with .env.prod and create .env.sandbox to match environments

artursudnik: or, maybe we should just leave one .env.example as a template and remove all others?

JGiter: it will be invonvenient. For example if you want to debug `prod`, then you need to pick `prod` values from `.env.example` and put them in `.env`. Simplre just copy `.env.prod` to `.env`

artursudnik: I see. I was thinking more like preparing env vars for deployments, not connecting to production or staging infrastructure.

jrhender (commented):
@artursudnik what does this look like if merging into `sandbox` branch rather than `develop`? Maybe we need to merge `develop` into `sandbox` before we would merge this PR into `sandbox`. I think trying it in Sandbox would be good, as `develop` does get some attention if it goes down

artursudnik:
@jrhender 
> what does this look like if merging into sandbox branch rather than develop

Adding this one inexpensive step for safety is always a good idea.

nichonien (approved):

jrhender (approved):
Looks good to me, thanks @artursudnik

nichonien (approved):

PR #530, opened 2022-11-02
feat: removing hardcoded keys pair from the docker image

artursudnik:


jrhender:
I think we need to add sealed secrets with the JWT keys as a part of this PR for the keys to be available to app: https://github.com/energywebfoundation/ssi-hub/blob/51acd9d285a10ecf266bcb85609c8f27b4bf9d52/devops/dev/values.yaml#L155

Does this sound right?

artursudnik:
@jrhender I tried to put keys into variables but it is problematic:
1. it is not possible to have multiple lines in them
2. even when having .pem file converted  one line triggers errors in the ssi-hub

So, after some research, I discovered that it is possible to mount a sealed secret into a container as a file, so this is a way we need to try.

artursudnik:
It is confirmed with DevOps that they are able to mount sealed secrets as a file into a container.

artursudnik:
The DevOps team prepared what is needed and this change has been already tested on Sandbox.

jrhender (approved):

PR #532, opened 2022-11-04
feat: configuration validation

artursudnik:
This PR:
- adds validations for env variables so that all required needs to be provided and all are validated formally
- adds a requirement that all env vars accessed with the ConfigService need to be defined in the [validation schema](https://github.com/energywebfoundation/ssi-hub/blob/feat/configuration-validation/src/env-vars-validation-schema.ts), otherwise they are stripped out


artursudnik:
Closing this PR due to serious problems with E2E testing revealed after adding env vars validation.

PR #536, opened 2022-11-09
feat: configuration validation

artursudnik:
This PR:
- adds validations for env variables so that all required needs to be provided and all are validated formally
- adds a requirement that all env vars accessed with the ConfigService need to be defined in the [validation schema](https://github.com/energywebfoundation/ssi-hub/blob/feat/configuration-validation/src/env-vars-validation-schema.ts), otherwise they are stripped out

TODO:
- [x] solve an issue that causes E2E tests to fail after validations are added. This is caused by the `process.env.DID_REGISTRY_ADDRESS` value overwritten [here](https://github.com/energywebfoundation/ssi-hub/blob/c7191fd73778665fb9b26a36b0b0e1a5af8a2e33/e2e/app.e2e.spec.ts#L39) after validations are executed. With validations enabled ConfigService caches validated values so that this leads to inconsistency between `process.env.DID_REGISTRY_ADDRESS` and `configService.get('DID_REGISTRY_ADDRESS')`

### DIFF
--- a/e2e/app.e2e.spec.ts
+++ b/e2e/app.e2e.spec.ts
@@ -18,7 +18,7 @@ import { ConfigService } from '@nestjs/config';
 
 export let app: INestApplication;
 
-jest.setTimeout(20000000);
+jest.setTimeout(600_000);
 describe('iam-cache-server E2E tests', () => {
   const provider = ethers.provider;
   const deployer = provider.getSigner(0);
--- a/e2e/app.e2e.spec.ts
+++ b/e2e/app.e2e.spec.ts
@@ -14,6 +14,7 @@ import { EthereumDIDRegistry } from '../src/ethers/EthereumDIDRegistry';
 import { EthereumDIDRegistry__factory } from '../src/ethers/factories/EthereumDIDRegistry__factory';
 import { didModuleTestSuite } from './did/did-service';
 import { Provider } from '../src/common/provider';
+import { ConfigService } from '@nestjs/config';
 
 export let app: INestApplication;
 
@@ -37,6 +38,7 @@ describe('iam-cache-server E2E tests', () => {
 
     didRegistry = await loadFixture(deployDidRegistry);
     process.env.DID_REGISTRY_ADDRESS = didRegistry.address;
+    const configService = new ConfigService(); // this is necessary to have values updated in the ConfigService provider
 
     const testingModule = await Test.createTestingModule({
       imports: [AppModule],
@@ -45,6 +47,8 @@ describe('iam-cache-server E2E tests', () => {
       .useValue(await spawnIpfsDaemon())
       .overrideProvider(Provider)
       .useValue(provider)
+      .overrideProvider(ConfigService)
+      .useValue(configService)
       .compile();
     app = testingModule.createNestApplication();
     appConfig(app);
--- a/src/app.module.ts
+++ b/src/app.module.ts
@@ -1,4 +1,4 @@
-import { Module } from '@nestjs/common';
+import { DynamicModule, Module } from '@nestjs/common';
 import { HttpModule } from '@nestjs/axios';
 import { ScheduleModule } from '@nestjs/schedule';
 import { ConfigModule, ConfigService } from '@nestjs/config';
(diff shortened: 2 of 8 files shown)


src/app.module.ts
```diff
@@ -27,6 +27,29 @@ import { HealthCheckModule } from './health-check/health-check.module';
...
+
+const validationOptions: ValidationOptions = {
+  stripUnknown: true,
+  allowUnknown: false,
+  abortEarly: false,
+};
+
+let config: DynamicModule;
+
+try {
```
JGiter, line 41: what is the reason of creating config object in try block? doesn't nests.js terminates app if confing invalid? will env variables be loaded outside of @Module?

src/app.module.ts
```diff
@@ -27,6 +27,29 @@ import { HealthCheckModule } from './health-check/health-check.module';
...
+  config = ConfigModule.forRoot({
+    isGlobal: true,
+    // below is a workaround for E2E not executing because of https://github.com/energywebfoundation/ssi-hub/blob/c7191fd73778665fb9b26a36b0b0e1a5af8a2e33/e2e/app.e2e.spec.ts#L39
+    validationOptions: !process.env.E2E ? validationOptions : null,
```
JGiter, line 45: skipping validation in tests reduces introduces potential vulnerabilities
also we would better be more specific in which enironment validation is enabled
```suggestion
    validationOptions: process.env.PROD ? validationOptions : null,
```

src/app.module.ts
artursudnik: It allows exiting with a clean message. This is the only reason I have implemented it like this.

src/app.module.ts
artursudnik: @JGiter 
I agree. However I would not call this vulnerabilities because this is not a validation of user input, but env variables. This is a workaround for the problem in E2E tests triggered by validation enabled. If we find another way to fix it, we can get rid of this workaround.

e2e/app.e2e.spec.ts
```diff
@@ -45,6 +47,8 @@ describe('iam-cache-server E2E tests', () => {
       .useValue(await spawnIpfsDaemon())
       .overrideProvider(Provider)
       .useValue(provider)
+      .overrideProvider(ConfigService)
+      .useValue(configService)
```
JGiter, line 51: Could you please give more details why to disable validation in tests? For me validation is passing without overriding ConfigService here and having `DID_REGISTRY_ADDRESSS` in `.env`

JGiter: Nevermind, I can see `"DID_REGISTRY_ADDRESS" is required`. This supposdly means that only values from `.env` are validated. May be we can use custom validation function https://docs.nestjs.com/techniques/configuration#custom-validate-function, which use values from both process and .env?

JGiter: So `DID_REGISTRY_ADDRESS` isn't found, because validation happens at the time, when test file imports `app.module` https://github.com/energywebfoundation/ssi-hub/blob/b8eaa5f3f29048c7c1640664fc41da78a30ff00d/e2e/app.e2e.spec.ts#L8 instead of when application is created. I think this is a good reason to move ConfigModule.forRoot in @Module

artursudnik: It behaves exactly the same when moved to the `@module`. I've checked this already. `ConfigModule.forRoot` is executed before the decorator.

artursudnik: The `ConfigService` does this when instantiated, so here we have the `process.env` object populated with them, but the problem was caused by validation added to the ConfigService. As a result, ConfigService creates its own copy of the `process.env` values before `process.env.DID_REGISTRY_ADDRESS = didRegistry.address;` is executed so manipulating them here leads to inconsistencies. This is the reason why providing another instance of it helped.

I already tried to provide a custom `validate` function to `ConfigModule.forRoot` but it behaves the same.

src/app.module.ts
artursudnik: It is 
```
Error: Config validation error: "PUBLIC_RESOLVER_ADDRESS" must be a valid ethereum address
exiting
```
vs
```
/Users/artur/WebstormProjects/energyweb/iam/ssi-hub/node_modules/@nestjs/config/dist/config.module.js:78
                throw new Error(`Config validation error: ${error.message}`);
                      ^
Error: Config validation error: "PUBLIC_RESOLVER_ADDRESS" must be a valid ethereum address
    at Function.forRoot (/Users/artur/WebstormProjects/energyweb/iam/ssi-hub/node_modules/@nestjs/config/dist/config.module.js:78:23)
    at Object.<anonymous> (/app.module.ts:63:18)
    at Module._compile (node:internal/modules/cjs/loader:1101:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1153:10)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Module.require (node:internal/modules/cjs/loader:1005:19)
    at require (node:internal/modules/cjs/helpers:102:18)
    at Object.<anonymous> (/main.ts:2:1)
    at Module._compile (node:internal/modules/cjs/loader:1101:14)
```

JGiter (approved):

nichonien (approved):

jrhender (approved):
Looks good, though maybe better to merge into `sandbox` first

artursudnik:
@jrhender it's already there: https://github.com/energywebfoundation/ssi-hub/commit/a5cf1661

jrhender:
> @jrhender it's already there: https://github.com/energywebfoundation/ssi-hub/commit/a5cf1661

Ah, thanks for reminding me @artursudnik !

PR #539, opened 2022-11-15
chore(helm): change JWT keys

michalziobro:


artursudnik (approved):

devops/dev/templates/jwt-keys-sealed-secret.yaml
```diff
@@ -6,11 +6,12 @@ metadata:
...
   template:
     metadata:
       creationTimestamp: null
       name: jwt-keys
       namespace: ics
+    type: Opaque
```
jrhender, line 16: @michalziobro what does this `Opaque` do?

michalziobro: @jrhender there are various types of secrets in k8s, Opaque is let's say a 'standard' secret:
https://stackoverflow.com/questions/49614439/kubernetes-secret-types

PR #541, opened 2022-11-16
fix: fixing /auth/status endpoint swagger definition 

artursudnik:
This PR fixes /auth/status endpoint is not marked with a padlock icon in the Swagger doc.

### DIFF
--- a/src/modules/auth/login.controller.ts
+++ b/src/modules/auth/login.controller.ts
@@ -9,7 +9,7 @@ import {
   UseGuards,
 } from '@nestjs/common';
 import { ConfigService } from '@nestjs/config';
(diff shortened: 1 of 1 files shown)


jrhender (commented):
@artursudnik without this change, does the `auth/status` endpoint not work in Swagger?

artursudnik:
@jrhender Without this, Swagger does not send an `Authorization` header. The endpoint works, but it is impossible to request it from the Swagger page to see this.

jrhender (approved):
Make sense, thanks

JGiter (commented):
Could you clarify which bug this PR fixes?

Thanks, I can see your answer above

JGiter (approved):

nichonien (approved):

PR #542, opened 2022-11-16
chore(helm): add JWT keys to other envs

michalziobro:


artursudnik (approved):

devops/prod/values.yaml
```diff
@@ -83,6 +83,8 @@ iam-cache-server-helm:
   config:
     secretRefName: ics-secrets
     configRefName: ics-cm
+    jwtKeysSecretName: jwt-keys
+    jwtKeysMountPath: "/keys"
```
jrhender, line 87: Ah, I think I am understanding better now. So jwtKeysSecretName tells kubernetes which secret to mount as files at `jwtKeyMountPath`. Is this right @michalziobro ? If so, how does Kubernetes know how to do this? Is it using the `template` file? If you have some documentation to link to, I don't mind reading that as well.

michalziobro: @jrhender exactly, so we create jwt-keys secret with bitnami sealed secrets and mount it to the directory provided. Documentation is here:
https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets-as-files-from-a-pod

and this is exactly what we do in our helm chart here:
https://github.com/energywebfoundation/iam-cache-server-helm/blob/7184ec677ae9d0cdd669bb8241599b011289233b/templates/deployment.yaml#L94

and here:
https://github.com/energywebfoundation/iam-cache-server-helm/blob/7184ec677ae9d0cdd669bb8241599b011289233b/templates/deployment.yaml#L35

jrhender: Makes sense, thanks for the explanation @michalziobro

jrhender (approved):

nichonien (approved):

PR #543, opened 2022-11-16
feat: add script to clear assets data db

nichonien:


jrhender (commented):
A couple questions:
- Are there any foreign keys which refer to the assets table? Is this a safe delete from that standpoint?
- How will this script be executed? Would it need to be run manually?

src/scripts/migrate_asset_data.js
```diff
@@ -0,0 +1,28 @@
...
+const { createConnection } = require('typeorm');
+const {
+  Asset,
+  AssetsHistory,
+} = require('../../dist/modules/assets/assets.entity.js');
+
+(async function () {
+  const connection = await createConnection({
+    // had to copy from orgmconfig because typeorm doesn't detect postgres driver
+    type: 'postgres',
```
artursudnik, line 11: I would get connection parameters from the `.env` file after adding `require('dotenv').config()` at the top of the script.

nichonien: > A couple questions:
> 
> * Are there any foreign keys which refer to the assets table? Is this a safe delete from that standpoint?

I don't think there is any other dependency on Asset and AssetsHistory, I will investigate more into this.

> * How will this script be executed? Would it need to be run manually?

Yes, this needs to called manually. I will add the command in `package.json`.

JGiter (approved):

JGiter (approved):

PR #546, opened 2022-11-24
feat: update assetmanager address

nichonien:
This PR updates `AssetManager` contract address to new `UUPS` deployed contract.


jrhender (approved):

jrhender (changes requested):
I think I would change the commit prefix. I don't think `feat` is warranted because there is no additional feature being added to ssi-hub. In my opinion, `chore` is more suitable.
This is important because it affects the version and changelog when releasing

artursudnik:
Shouldn't this occurrence also be changed? https://github.com/energywebfoundation/ssi-hub/blob/68cc37c1cc36c32dafb3e8e08688fdc46b27b827/.env.dev#L38-L38

artursudnik:
> @artursudnik I think I have updated this as well, can you double check please?

oh, I apologize. I need another coffee ;)

nichonien:
@jrhender Kindly review this, thanks!

JGiter (approved):
Probably should be ok, we can though consider complete environment separation or a least separate dev/sandbox from staging/prod

nichonien:
> Probably should be ok, we can though consider complete environment separation or a least separate dev/sandbox from staging/prod

Hmm, are you suggesting to have these environments use there own / different deployed contract?

JGiter:
> > Probably should be ok, we can though consider complete environment separation or a least separate dev/sandbox from staging/prod
> 
> Hmm, are you suggesting to have these environments use there own / different deployed contract?

Yes, staging is semi-prod, so it should be not affected by bugs in dev

jrhender (approved):
Looks good, thanks @nichonien

PR #552, opened 2023-01-02
feat: adding log messages

artursudnik:
This PR:
- adds application error log messages for failed processing of queued jobs
- handles claims pinning errors by writing an application error log message

### DIFF
--- a/src/modules/did/did.processor.ts
+++ b/src/modules/did/did.processor.ts
@@ -44,7 +44,11 @@ export class DIDProcessor {
       this.logger.debug(`processing cache add for ${job.data}`);
       const doc = await this.didService.addCachedDocument(job.data);
 
-      await this.pinClaims(doc.service.map((s) => s.serviceEndpoint));
+      await this.pinClaims(doc.service.map((s) => s.serviceEndpoint)).catch(
+        (err) => {
+          this.logger.error(`error pinning the claim: ${err}`);
+        }
+      );
     } catch (err) {
       this.logger.error(`error adding ${job.data}: ${err}`);
       throw err;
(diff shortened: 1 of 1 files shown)


src/modules/did/did.processor.ts
```diff
@@ -40,23 +40,41 @@ export class DIDProcessor {
...
-    this.logger.debug(`processing cache add for ${job.data}`);
-    const doc = await this.didService.addCachedDocument(job.data);
+    try {
+      this.logger.debug(`processing cache add for ${job.data}`);
+      const doc = await this.didService.addCachedDocument(job.data);
 
-    await this.pinClaims(doc.service.map((s) => s.serviceEndpoint));
+      await this.pinClaims(doc.service.map((s) => s.serviceEndpoint)).catch(
+        (err) => {
+          this.logger.error(`error pinning the claim: ${err}`);
```
artursudnik, line 49: we need to decide if this error needs to be rethrown here

src/modules/did/did.processor.ts
```diff
@@ -40,23 +40,41 @@ export class DIDProcessor {
...
+      if (this.configService.get<boolean>('DID_SYNC_MODE_FULL')) {
+        doc = await this.didService.addCachedDocument(job.data, true);
+      } else {
+        doc = await this.didService.incrementalRefreshCachedDocument(job.data);
+      }
 
-    await this.pinClaims(doc.service.map((s) => s.serviceEndpoint));
+      await this.pinClaims(doc.service.map((s) => s.serviceEndpoint)).catch(
+        (err) => {
+          this.logger.error(`error pinning the claim: ${err}`);
```
artursudnik, line 71: we need to decide if this error needs to be rethrown here

src/modules/did/did.processor.ts
nichonien: This could be helpful.

JGiter (approved):

PR #553, opened 2023-01-10
task/SWTCH-1542 IPFS module

JGiter:
This PR is tested with development cluster and ipfs.infura.io. On average saving one claim in cluster takes 30 seconds
Noticed issues with development cluster:

- claims are not actually pinned. After claim is stored it can be found only during some period of time, which is no more than a day
- resolution of some claims takes so much time, that DidStore decides that they are not persisted and saves them again

Example logs after third login
image


src/modules/auth/login.strategy.ts
```diff
@@ -2,27 +2,29 @@ import { LoginStrategy } from 'passport-did-auth';
...
 
 @Injectable()
 export class AuthStrategy extends PassportStrategy(LoginStrategy, 'login') {
   constructor(
     configService: ConfigService,
-    @Inject('IPFSClientConfig') ipfsConfig
+    @Inject(IPFSGatewayConfigToken) ipfsConfig: IpfsGatewayConfig
   ) {
-    let loginStrategyParams = {
+    let loginStrategyParams: Omit<LoginStrategyOptions, 'jwtSecret'> = {
```
JGiter, line 14: `jwtSecret` is not used by LoginStrategy

e2e/app.e2e.spec.ts
```diff
@@ -36,17 +39,17 @@ describe('iam-cache-server E2E tests', () => {
 
     didRegistry = await loadFixture(deployDidRegistry);
     process.env.DID_REGISTRY_ADDRESS = didRegistry.address;
-    process.env.IPFS_CLUSTER_ROOT = 'http://localhost:8080';
-    process.env.IPFS_CLUSTER_USER = 'not-required-locally';
-    process.env.IPFS_CLUSTER_PASSWORD = 'not-required-locally';
```
jrhender, line 41: Aren't the Cluster User and Password need to authenticate to the cluster when not running locally?

src/modules/auth/login.strategy.ts
jrhender: Do you know why this is?

e2e/app.e2e.spec.ts
JGiter: No, they not needed. Initially this variables were required for environment validation. I made them optional and we don't have to specify empty strings now

src/modules/auth/login.strategy.ts
JGiter: Sorry, I was not accurate. `jwtSecret` is required by LoginStrategy to encode logged in user, but we decided to login to ssi-hub without encoding and because of this don't passing jwtSecret

src/modules/ipfs/ipfs.service.ts
```diff
@@ -26,4 +54,48 @@ export class IPFSService {
...
+   * @returns Stringified credential
+   */
+  public async get(cid: string): Promise<string> {
+    let claim: string;
+    const getFromCluster = this.didStoreCluster.get(cid);
+    const getFromInfura = this.didStoreInfura.get(cid);
+    try {
+      claim = await Promise.race([getFromCluster, getFromInfura]);
+    } catch (e) {
+      try {
```
JGiter, line 71: If `try` failed because of getting from cluster, try to get from Infura

.gitignore
```diff
@@ -2,7 +2,6 @@
 /dist
 /node_modules
 /src/ethers
-.env
```
jrhender, line 5: Why removing the ignore for .env? Couldn't this lead to folks checking in their .env files accidentally?

e2e/app.e2e.spec.ts
```diff
@@ -32,21 +36,22 @@ describe('iam-cache-server E2E tests', () => {
   }
 
   beforeAll(async () => {
+    await import('../src/scripts/truncate-db');
```
jrhender, line 39: It seems that just importing the script has the side-effect of executing the script then? Is this to avoid needing to call a function after the import?

jrhender (commented):
I'm still digesting the PR but I added a couple questions to start

.gitignore
JGiter: sorry, I forget to comment this. .env already ignored in line 40

src/env-vars-validation-schema.ts
```diff
@@ -89,14 +89,15 @@ export const envVarsValidationSchema = Joi.object({
   ASSETS_SYNC_HISTORY_INTERVAL_IN_HOURS: Joi.number().positive().required(),
   ASSETS_SYNC_ENABLED: Joi.boolean().required(),
 
+  IPFS_CLIENT_URL: Joi.string().uri(),
+  IPFS_CLIENT_PROTO: Joi.string(),
```
JGiter, line 93: we will need to change `IPFS_PROTOCOL` in [gp4btc](https://github.com/energywebfoundation/gp4btc/blob/bdd5dd75b8172ef62e15025e2ea7096851c7ae7a/devops/gp4btc-ics-prod/values.yaml#L121) to `IPFS_CLIENT_PROTO`

e2e/app.e2e.spec.ts
JGiter: Not exactly. I didn't define truncate-db as exported function, but as executable file, because I want to call it as npm script. So this dynamic import is price for not creating another file which only allows to import truncate-db. May be its not really good justification

.gitignore
```diff
@@ -41,6 +41,7 @@ ormlogs.log
 
 # Using either dev or prod docker-compose file
 docker-compose.yml
+!/e2e/setupIpfsCluster/docker-compose.yml
```
jrhender, line 44: Minor point, but wouldn't it make sense to have this below the other line that was added? Why is this in a different section of the file?

devops/prod/values.yaml
```diff
@@ -182,4 +183,4 @@ iam-cache-server-helm:
...
\ No newline at end of file
+      IPFS_CLUSTER_PASSWORD: AgCncDxtjXgBmdaMT+a7NN8PIZQaNH7PJ/C5RmGkcXwYUsS9JONjWqBny+7OEPM9Y82DaYGrsH3LwBrCG2wJ8ja5zelHfrkl4ue9HnplpbhnG9eigMRer0pFHbwdrqagGZNI5EOZXEQNZb3uy/BUFbvB+dWvEbYnNz4UYHU/3hQAAbuQ81VniTRmmSb6kbNN58mHVHCwbC1kjQsssQ1cVE5aTSLZFfxgb1zIvgrhZKKWmG0ygZDV78XMHZ8fCvBACcIlt9Suy9JdpJMZG8omvtYwF6wvGj7w1DiZwr67Pmt0i2959MgFA6nIYxo8mJbEaDaAnTYAheiNHWnTXxVXCQDECN5xNkd81tNfj0wMQa8CqgH0KbH/lJCFSyruKA4Up/7h7xhU/SG9nx8DyzV1g0phxkggHcXBeFGeupwHS6d5O80N9ThxVMLznrHgpyqdz4ZZORtGjvPoH+3ehLg22CbZasPhEPzkfT4wg0NBRNqf7YUqEz+WS41CQVHL/PqON39YzEkqy0MZhAdhSWKjSopdQBfZxji5DujbofX17qSMD3IWsTpF3IV3pRCZAqAmH77R93FxJWBtrJiUztGf4Po8fiYScusNiEORhFru0x1+avH8aHxbNvDfsPGN5NOAsoQ8AD6SAvPDwTacl8YydccAv71To+sNQYP+qgmKk4ir6Nn5sB/tAXDLzvfx9QHU+j6jgY48abU4q2ZKA2GKG6po/lu+jawugA==
```
jrhender, line 186: Another minor point, but probably this line shouldn't be changed. It seems there is no actual change, but maybe better not to take a risk and have an extra entry in the git history for this line.

src/modules/ipfs/ipfs.controller.ts
```diff
@@ -0,0 +1,46 @@
...
+@UseInterceptors(SentryErrorInterceptor)
+@Controller({ path: 'ipfs', version: '1' })
+export class IPFSController {
+  constructor(private ipfsService: IPFSService) {}
+
+  @Get('/:cid')
+  @ApiTags('IPFS')
+  @ApiOperation({
+    summary: 'Returns credential from IPFS Store',
+    description: 'Returns credential represented by service in DID document.',
```
jrhender, line 26: I think it's better to have this more generic. It's not really specific to credentials, so no need to limit it to that, even if that is what we are currently using it for.
```suggestion
    summary: 'Returns content from IPFS Store',
    description: 'Returns data identified by the provided content identifier (cid). This can be used, for example, to resolve credentials linked in the service endpoint of a DID document.',
```

src/modules/ipfs/ipfs.controller.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  @Post()
+  @ApiTags('IPFS')
+  @ApiBody({
+    type: 'string',
+    description: 'Stringified credential',
+  })
+  @ApiOperation({
+    summary: 'Saves credential in IPFS',
+    description: 'Saves credential on IPFS and returns its CID',
```
jrhender, line 41: As with above comment, I think it would be better if the description was generic, though we would use it for credentials.
```suggestion
    summary: 'Saves content in IPFS',
    description: 'Saves content in IPFS and returns its CID',
```

src/modules/ipfs/ipfs.controller.ts
```diff
@@ -0,0 +1,46 @@
...
+  @ApiParam({ name: 'cid', type: 'string', required: true })
+  public async get(@Param('cid', CIDPipe) cid: CID): Promise<string> {
+    return this.ipfsService.get(cid.toString());
+  }
+
+  @Post()
+  @ApiTags('IPFS')
+  @ApiBody({
+    type: 'string',
+    description: 'Stringified credential',
```
jrhender, line 37: I think this isn't accurate based the current API. The current API is accepting a `Record`. Not sure if the solution is to change the API or the documentation. I think accepting a string might be more general than a `Record` and so might be better.

jrhender (changes requested):
I think the documentation in `ipfs.controller` needs to be tweaked

src/modules/ipfs/ipfs.module.ts
```diff
@@ -1,38 +1,109 @@
...
-    const IPFS_CLIENT_PORT = config.get<string>('IPFS_CLIENT_PORT');
+const IPFSInfuraConfigProvider = {
+  provide: IPFSInfuraConfigToken,
+  useFactory: (config: ConfigService): IpfsInfuraConfig => {
+    const IPFS_CLIENT_URL = config.get<string>('IPFS_CLIENT_URL');
+    const IPFS_CLIENT_PROTOCOL = config.get<string>('IPFS_CLIENT_PROTOCOL');
     const IPFS_CLIENT_HOST = config.get<string>('IPFS_CLIENT_HOST');
+    const IPFS_CLIENT_PORT = config.get<string>('IPFS_CLIENT_PORT');
+    const ipfsConfig: IpfsInfuraConfig = {
+      url: IPFS_CLIENT_URL,
```
jrhender, line 25: Why was `url` introduced in `IpfsInfuraConfig`? It seems a URL could be derived from `protocol`, `host` and `port`, so it seems to me like it might conflict with those values

src/modules/ipfs/ipfs.service.ts
jrhender: Could this be simplified by using [Promise.any](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/any)? With this, I think this second `try` wouldn't be needed.

src/modules/ipfs/ipfs.module.ts
jrhender: Ah yes, I see from the documentation that `url` supersedes the other values: https://github.com/ipfs/js-ipfs/tree/master/packages/ipfs-http-client#createoptions

src/modules/ipfs/ipfs.service.ts
JGiter: Thanks. I didn't know about this
It is also requires ES2021 typescript lib

devops/prod/values.yaml
JGiter: Probably line ending here was changed. I tried to reset this file to develop but it didn't help. I added .gitiattributes to prevent such issues

src/modules/ipfs/ipfs.controller.ts
JGiter: Ok, I changed content to string

src/modules/ipfs/ipfs.service.ts
```diff
@@ -26,4 +54,45 @@ export class IPFSService {
...
+  public async save(credential: string): Promise<string> {
+    try {
+      return this.didStoreCluster.save(credential);
+    } catch (_) {
+      this.logger.warn(
+        `Error saving ${credential} in cluster. Saving in Infura`
+      );
+      return this.didStoreInfura.save(credential);
+    }
+  }
```
jrhender, line 97: @JGiter I wonder if we should continue to save in Infura as well as the cluster. 

In the PR description you mentioned that:
> claims are not actually pinned. After claim is stored it can be found only during some period of time, which is no more than a day

I am concerned that if we are pinning to cluster but the cluster persistence isn't reliable, then we may save the data but it will not reliably be available. What do you think?

JGiter: Yes, it make sense

e2e/setupIpfsCluster/index.ts
```diff
@@ -0,0 +1,53 @@
...
+      }
+    }
+  );
+
+  try {
+    await waitOn({
+      // https://ipfscluster.io/documentation/reference/api/
+      resources: ['http-get://localhost:8080/id'],
+      delay: 5000,
+      // if your internet connection is slow try to increase timeout to be able to download all images
```
jrhender, line 29: I think this line isn't relevant because `localhost` calls don't go over the internet
```suggestion
```

devops/prod/values.yaml
jrhender: Ok, thanks

e2e/setupIpfsCluster/index.ts
JGiter: Actually the delay in spinning cluster can be caused by internet calls. Cluster need to download `nginx` and `ipfs` images listed in docker-compose

jrhender: Ah gotcha, I see what you mean by `images` now

src/modules/ipfs/ipfs.service.ts
```diff
@@ -26,4 +44,53 @@ export class IPFSService {
...
+  public async save(credential: string): Promise<string> {
+    const [clusterCID, infuraCID] = await Promise.allSettled([
+      this.didStoreCluster.save(credential),
+      this.didStoreInfura.save(credential),
+    ]);
+    if (clusterCID.status === 'fulfilled') {
+      return clusterCID.value;
+    } else if (infuraCID.status === 'fulfilled') {
+      this.logger.warn(
+        `Error saving ${credential} in cluster. Saving in Infura`
```
jrhender, line 87: A bit more precise
```suggestion
        `Error saving ${credential} in cluster. Was saved to Infura as backup`
```

devops/prod/values.yaml
```diff
@@ -135,6 +135,7 @@ iam-cache-server-helm:
     LOGS_DIRECTORY: /app/logs
     NESTJS_PORT: 3000
     SENTRY_ENV: prod
+    IPFS_CLIENT_PROTOCOL: https
```
jrhender, line 138: @JGiter will we also need to set this for the gp4btc instances?

jrhender (commented):
PR is looking good. Added a couple comments, but I think we could merge if we are ok with the cluster pinning not working currently.

jrhender (approved):
Looks good to me, let's give it a try.

PR #554, opened 2023-01-11
feat(role.dto): add optional schema prop

jrhender:
- [x] add prop to dto
- [x] test with role def that has schema

I tested locally that `schema` is returned from a role that issuerField/requestorField with it:
image


artursudnik:
I see a typo on screenshot: `isserTestSchema` is this generated or just entered by a human?

nichonien (approved):

jrhender:
> I see a typo on screenshot: `isserTestSchema` is this generated or just entered by a human?

@artursudnik just entered by a human (me 😆 )

PR #562, opened 2023-01-31
task/SWTCH-1983 restrict cross-request origins

JGiter:
List of changes
1. Adds options to restrict origin of login request. When `RESTRICT_CORS_ORIGIN` environment variable is set to `true` only origins defined in `ALLOWED_ORIGINS` environment variable will be able to login. `ALLOWED_ORIGINS` are specified as comma separated values
2. Adds check that matches request origin to origin of `refresh_token`

If login request has no origin, then it is authenticates regardless of `RESTRICT_CORS_ORIGIN`


src/db/cli.ts
```diff
@@ -12,4 +12,6 @@ export const dataSource = new DataSource({
     get: (key: string, def?: string) => process.env[key] || def,
   }),
   type: 'postgres',
+  installExtensions: true,
```
JGiter, line 15: Attempt to fix https://github.com/energywebfoundation/ssi-hub/actions/runs/4066246957/jobs/7002070873#step:11:161

e2e/claim/claim-with-request.testSuite.ts
```diff
@@ -315,29 +318,8 @@ export const claimWithRequestTestSuite = () => {
     expect(body.redirectUri).toBe(requestOrigin);
   });
 
-  it(`should throw an error when request origin is invalid`, async () => {
```
JGiter, line 318: removed this bc invalid origin will not be able reach RoleClaim constructor now since origin is checked during authentication

src/db/cli.ts
artursudnik: is this still necessary?

artursudnik: oh, nevermind. it's outdated :)

JGiter: sorry, forget to rm comment. will rm now

src/modules/auth/token.service.ts
```diff
@@ -95,22 +99,33 @@ export class TokenService {
...
+      origin === req.headers['origin'];
+    const isServerRequestOrGETFromSameDomain =
+      req.headers['origin'] === undefined;
+    if (
+      !isBrowserRequestFromAuthenticatedOrigin &&
+      !isServerRequestOrGETFromSameDomain
+    ) {
+      throw new UnauthorizedException(
+        `Token origin ${origin} does not matches request origin ${req.headers['origin']}`
+      );
```
jrhender, line 128: As I think this is a bit separate from the direct CORS configuration and is a refactoring rather than new functionality, I think it would be better to have this in a separate PR. We can leave it here, but just a point to keep in mind for future.

jrhender (changes requested):
Nice work on the test cases 👍 

I think the most important change is that I think it would be better to use the `cors` package (via Nest config) rather than rolling our own origin check.

JGiter:
> I think the most important change is that I think it would be better to use the `cors` package (via Nest config) rather than rolling our own origin check.

I added custom middleware because from the first look at cors configuration options I didn't see a way to pass requests without `Origin` header, but now I see that it can be done dynamically. I'll try this

src/env-vars-validation-schema.ts
```diff
@@ -84,6 +107,7 @@ export const envVarsValidationSchema = Joi.object({
   JWT_REFRESH_TOKEN_EXPIRES_IN: Joi.string().required(), // TODO: implement validation of this flexible format
   JWT_ACCESS_TOKEN_NAME: Joi.string().required(),
   JWT_REFRESH_TOKEN_NAME: Joi.string().required(),
+  ALLOWED_ORIGINS: Joi.stringifiedArray(),
```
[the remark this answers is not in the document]
JGiter: I tried to verify `ALLOWED_ORIGINS` as array, but it failed. I assume that all values in `.env` are strings. I can't cite authoritative source for this, but here related discussion https://stackoverflow.com/questions/31552125/defining-an-array-as-an-environment-variable-in-node-js. The point of interest is the second answer

e2e/auth/login.testSuite.ts
```diff
@@ -54,6 +65,59 @@ export const authLoginTestSuite = () => {
...
+      const identityToken = await getIdentityToken(provider, wallet);
+      await request(app.getHttpServer())
+        .post('/v1/login')
+        .set('origin', allowedOrigins[0])
+        .send({
+          identityToken,
+        })
+        .expect(201);
+    });
+
```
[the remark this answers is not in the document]
JGiter: We have one test when origin header isn't set https://github.com/energywebfoundation/ssi-hub/blob/44b16b3694f27773b0ea9ba02b26e6fe295ba657/e2e/auth/login.testSuite.ts#L35. But I agree that some more can be added. Grouping tests when origin is set or not will also help.
By the way from my opinion request to `v1/search/test` in your test should fail, because `Origin` header doesn't match origin of access token, which undefined. Why we should allow this request?

src/modules/auth/auth.module.ts
```diff
@@ -46,12 +49,28 @@ import { STATUS_LIST_MODULE_PATH } from '../status-list/status-list.const';
...
+  }
 
   configure(consumer: MiddlewareConsumer) {
     consumer
-      .apply((req, res, next) =>
-        this.tokenService.handleOriginCheck(req, res, next)
+      .apply((req: Request, res: Response, next: NextFunction) =>
+        this.checkAllowedOrigin(req, next)
+      )
+      .forRoutes({ path: '/*', method: RequestMethod.ALL })
```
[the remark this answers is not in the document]
JGiter: I agree. Though `app.enableCors` is more consistent with how we adding other middleware, then `NestFactory.create`. We are calling methods on `app`

src/modules/auth/token.service.ts
JGiter: Yes, it is not directly relates to CORS. I separated matching origin to token in order not to repeat it separately for access and refresh token

e2e/auth/login.testSuite.ts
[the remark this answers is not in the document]
jrhender: Ah, thanks, I didn't see the test that you linked to. I think that test is probably enough. In fact, the test you linked to is better because it is retrieving the token from the cookie header in the response and not from the response body (see https://energyweb.atlassian.net/browse/SWTCH-2020)

src/modules/auth/auth.module.ts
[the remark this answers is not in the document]
jrhender: Ah okay, I didn't realize that `app.enableCors` and the `cors` object in `NestFactory.create` are the same thing but that makes sense. You're right, it's better to modify the line that already has `app.enableCors` rather than switch the pattern we're using https://github.com/energywebfoundation/ssi-hub/blob/605820d672392b0ee72c107c2d150aef39d9b4c1/src/main.ts#L73

src/env-vars-validation-schema.ts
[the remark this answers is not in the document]
jrhender: Ah makes sense, thanks for the explanation 👍

jrhender (commented):
@JGiter how hard would it be to add an e2e test for the new "allow all origin" case? Is it hard to have an e2e test with different config?

src/common/test.utils.ts
```diff
@@ -14,6 +17,36 @@ export function appConfig(app: INestApplication) {
...
+                new UnauthorizedException(`Origin ${origin} is not allowed`),
+                false
+              );
+            }
+          } else {
+            callback(undefined, true);
+          }
+        }
+      : true,
+  });
```
jrhender, line 49: It occurred to me that this code is duplicating the code in `main.ts`. Maybe we can do it in a separate PR, but is there a way to remove this duplication? Do you think this is worth doing?

JGiter:
> @JGiter how hard would it be to add a test for the new "allow all origin" case? Is it hard to have a test with different config?

Unfortunately I don't see a simple way to change environment variable `RESTRICT_CROSS_ORIGINS` per test. There are two challenges:
- in order to change CORS configuration application has to be relaunched. So we have to start/stop application in each test, which results in longer CI
- CORS configuration have to be done in top level `describe`, where application is started, but inner `describe` has no way to hook into top-level
Nevertheless I added test for login with reflected origin, which is skipped by default (when `RESTRICT_CROSS_ORIGINS` is true), but can be enabled when all origins allowed https://github.com/energywebfoundation/ssi-hub/blob/1c42180a15d8331a6e76cb25499a8965cdf8becb/e2e/auth/login.testSuite.ts#L159

src/common/test.utils.ts
JGiter: Yes, this duplication should be removed

jrhender (commented):
It looks good, thanks @JGiter . Before merging, let's coordinate with the DDHub team so that they are ready to test when we merge.

jrhender:
Though the PR looks good and I think could be merged, I will add the "Do Not Merge" while we coordinate the deployment of this change with the DDHub change and the wider AEMO community.

@JGiter To help with communicating this change, could you add a description to this PR please?

devops/staging/values.yaml
```diff
@@ -117,6 +117,8 @@ iam-cache-server-helm:
     STRATEGY_NUM_BLOCKS_BACK: 10
     JWT_ACCESS_TOKEN_EXPIRES_IN: 15m
     JWT_REFRESH_TOKEN_EXPIRES_IN: 1h
+    RESTRICT_CORS_ORIGINS: true
+    ALLOWED_ORIGINS: https://gp4btc-ui-stg.energyweb.org/,https://gp4btc-explorer-stg.energyweb.org/
```
jrhender, line 121: As we now have the deployment for gp4btc elsewhere, the only ui that we need access the "staging" environment is switchboard-staging
```suggestion
    RESTRICT_CORS_ORIGINS: true
    ALLOWED_ORIGINS: https://switchboard-staging.energyweb.org
```

jrhender (commented):
A small comment about adjusting `staging` values now that we have separate gp4btc environment

jrhender:
@JGiter FYI, I added prod values (https://github.com/energywebfoundation/ssi-hub/pull/562/commits/d47474982e895f15da6ec36f6b0e23571771e5ff) so that we could potentially merge https://github.com/energywebfoundation/ssi-hub/pull/576

jrhender (approved):

PR #565, opened 2023-02-02
feat: switch to nginx ingress

Timtech4u:
Once merged, this needs to be synced with lower envs


jrhender:
@Timtech4u shall we close this PR?

PR #568, opened 2023-02-02
chore(CI): completely rewriting the `Build and test` workflow

artursudnik:
This PR redesigns the `Build and test` GH workflow so that:
- dependencies, application build, Unit tests, and E2E tests are executed separately
- Unit tests and E2E tests are executed in parallel and because their executions are long this speeds up a workflow execution
- DB schema migrations are executed explicitly before any application code is executed
image

It also adds a workflow that prepopulates caches so that PRs based on the `develop` branch, do not waste resources for the first installation of dependencies.


.github/workflows/build-test.yml
```diff
@@ -4,12 +4,122 @@ name: Build and test
...
+        with:
+          # npm cache files are stored in `~/.npm` on Linux/macOS
+          path: |
+            ~/.npm
+          key: ${{ runner.os }}-npm-${{ hashFiles('package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-npm-${{ hashFiles('package-lock.json') }}
+            ${{ runner.os }}-npm-
+      - name: Cache node_modules
+        id: cache-npm
```
JGiter, line 78: `cache-node-modules` looks more precise name
```suggestion
        id: cache-node-modules
```

JGiter (approved):

nichonien (approved):

.github/workflows/build-test.yml
artursudnik: Yes, it needs to correspond to the step name. Thanks!

PR #573, opened 2023-02-08
fix: remove chrome specific cookies update

nichonien:
This PR removes chrome browser specific validation and cookies update

### DIFF
--- a/src/modules/auth/cookies.service.ts
+++ b/src/modules/auth/cookies.service.ts
@@ -1,6 +1,5 @@
 import { Injectable } from '@nestjs/common';
 import { CookieOptions } from 'express';
-import useragent from 'useragent';
 
 @Injectable()
 export class CookiesServices {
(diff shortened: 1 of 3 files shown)


src/modules/auth/cookies.service.ts
```diff
@@ -10,11 +9,6 @@ export class CookiesServices {
       sameSite: 'none',
       secure: true,
     };
-    const { family, major } = useragent.parse(userAgent) || {};
```
artursudnik, line 13: linter complains about `userAgent` not used :)

nichonien: @artursudnik Yes! I thought of removing the `userAgent` parameter, but then left it as is so that if any validation needs to be added we can use it. Do you think changing the signature would be good?

artursudnik: we can always add it back. My personal preference is to keep the code clean, to avoid confusion later.

JGiter (approved):

jrhender (approved):

PR #579, opened 2023-02-14
fix: instantiate cookiesservice in logincontroller

nichonien:
This PR instantiates `CookiesService` in `LoginController` and removes it from constructor params.

### DIFF
--- a/src/modules/auth/login.controller.ts
+++ b/src/modules/auth/login.controller.ts
@@ -20,9 +20,9 @@ import { RoleService } from '../role/role.service';
 @ApiTags('Auth')
 @Controller({ version: '1' })
 export class LoginController {
+  private cookiesServices = new CookiesServices();
   constructor(
     private tokenService: TokenService,
-    private cookiesServices: CookiesServices,
     private configService: ConfigService,
     private roleService: RoleService
   ) {}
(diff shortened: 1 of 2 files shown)


JGiter (commented):
@nichonien could you point out what is the benefirs of manual instantiating of `CookiesService`?

nichonien:
> @nichonien could you point out what is the benefirs of manual instantiating of `CookiesService`?

@JGiter There is no actual benefit, but honestly this instance coming from outside doesn't add any value or provide anything new with comparison to an internal instantiation (but eases the implementation by reducing constructor params). Secondly, this is an attempt to check if the `Snyk` code scan can pick the `security` property in `cookies` if it is created within the same class.

src/modules/auth/login.controller.ts
```diff
@@ -20,9 +20,9 @@ import { RoleService } from '../role/role.service';
 @ApiTags('Auth')
 @Controller({ version: '1' })
 export class LoginController {
+  private cookiesServices = new CookiesServices();
```
jrhender, line 23: I think this is unlikely to change snyk's results and we lose the benefits of nest Dependency Injection. I think that I would try copying the `CookiesService` code directly into the LoginController

jrhender (commented):
I'm at bit skeptical that this will work, but you never know.

Could you try running the Snyk cli against this branch to see if we can test it before merging? https://docs.snyk.io/snyk-cli/test-for-vulnerabilities/grouping-projects-by-branch-or-version

jrhender:
@nichonien shall we close this PR?

nichonien:
Closing as this is not needed anymore.

PR #581, opened 2023-02-16
feat: switch from ecr to gchr

Timtech4u:
The PR adopts GHCR while dropping ECR for the ssi-hub images.
It also updates the image repository used in the ArgoCD deployments for ssi-hub.

### DIFF
--- a/.github/workflows/build-deploy.yml
+++ b/.github/workflows/build-deploy.yml
@@ -148,7 +148,7 @@ jobs:
         with:
           context: .
           push: ${{  needs.configure.outputs.PUSH_DOCKER  }}
-          platforms: linux/amd64,linux/arm64
+          platforms: linux/amd64
           tags: ${{ steps.tags.outputs.TAGS }}
           labels: ${{ steps.meta.outputs.labels }}
           file: ./docker/prod/Dockerfile
--- a/.github/workflows/build-deploy.yml
+++ b/.github/workflows/build-deploy.yml
@@ -7,8 +7,12 @@ on:
       - 'devops/**'    
   workflow_dispatch:
 
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}
 
 jobs:
+
   cancel-previous:
     name: 'Cancel Previous Runs'
     runs-on: ubuntu-latest
@@ -17,7 +21,7 @@ jobs:
       - uses: styfle/cancel-workflow-action@0.10.0
         with:
           access_token: ${{ github.token }}
-
+          
   unique_id:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/build-deploy.yml
```diff
@@ -26,74 +30,147 @@ jobs:
...
-            echo "::set-output name=env_name::prod"
-          fi
-      - name: Use specified environment secrets from previous step
-        run: echo "Environment secrets set name ${{ steps.env_name.outputs.env_name }}"
+              echo "Running on branch ${{ github.ref }}"
+              if [ "${{ github.ref }}" = "refs/heads/develop" ]; then
+                echo "ENVIRONMENT=dev" >> $GITHUB_OUTPUT
+                echo "PRERELEASE=true" >> $GITHUB_OUTPUT
+                echo "IMAGE_TAG=$(uuidgen)" >> $GITHUB_OUTPUT
+                echo "TAG_DRY_RUN=true" >> $GITHUB_OUTPUT
```
jrhender, line 54: So we'll be doing dry runs on develop to start? And then once we're happy with it, we'll update to `false` here I suppose?

.github/workflows/build-deploy.yml
```diff
@@ -26,74 +30,147 @@ jobs:
...
+              
+      - name: Build and push Docker image
+        uses: docker/build-push-action@v3
+        with:
+          context: .
+          push: ${{  needs.configure.outputs.PUSH_DOCKER  }}
+          platforms: linux/amd64,linux/arm64
+          tags: ${{ steps.tags.outputs.TAGS }}
+          labels: ${{ steps.meta.outputs.labels }}
+          file: ./docker/Dockerfile
```
jrhender, line 154: This section needs to be tweaked as we've got two docker builds to perform, one from `docker/prod/Dockerfile` and one from `docker/sentry-upload/Dockerfile`

jrhender (changes requested):

.github/workflows/build-deploy.yml
```diff
@@ -26,74 +30,147 @@ jobs:
...
-          fi
-      - name: Use specified environment secrets from previous step
-        run: echo "Environment secrets set name ${{ steps.env_name.outputs.env_name }}"
+              echo "Running on branch ${{ github.ref }}"
+              if [ "${{ github.ref }}" = "refs/heads/develop" ]; then
+                echo "ENVIRONMENT=dev" >> $GITHUB_OUTPUT
+                echo "PRERELEASE=true" >> $GITHUB_OUTPUT
+                echo "IMAGE_TAG=$(uuidgen)" >> $GITHUB_OUTPUT
+                echo "TAG_DRY_RUN=true" >> $GITHUB_OUTPUT
+                echo "PUSH_DOCKER=false" >> $GITHUB_OUTPUT
```
jrhender, line 55: Why `false` here? I think we want to push for develop as well, we're just not adding the semvar tag, right?
```suggestion
                echo "PUSH_DOCKER=true" >> $GITHUB_OUTPUT
```

.github/workflows/build-deploy.yml
```diff
@@ -26,74 +30,147 @@ jobs:
...
+      - name: Bump version and push tag
+        id: tag_version
+        uses: mathieudutour/github-tag-action@v6.1
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          release_branches: release
+          pre_release_branches: master    # these branches will produce pre-release tag (eg. v0.1.2-beta.0)
+          append_to_pre_release_tag: beta
+          custom_release_rules: major:major:Major Changes,minor:minor:Minor Changes,chore:patch:Chores
+          dry_run: ${{  needs.configure.outputs.TAG_DRY_RUN  }}   # pushes to develop won't trigger tagging
```
jrhender, line 91: This looks good, my only concern with this is that the version in `package.json` won't reflect this tag & release version. Maybe this is okay, but curious if you are aware of a way to improve this.

.github/workflows/build-deploy.yml
```diff
@@ -26,74 +30,147 @@ jobs:
...
+              else
+                echo "TAGS=${{ steps.meta.outputs.tags }},${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{  needs.configure.outputs.IMAGE_TAG  }},${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{  needs.tag-release.outputs.tag  }}" >> $GITHUB_OUTPUT
+              fi
+              
+      - name: Build and push Docker image
+        uses: docker/build-push-action@v3
+        with:
+          context: .
+          push: ${{  needs.configure.outputs.PUSH_DOCKER  }}
+          platforms: linux/amd64,linux/arm64
```
jrhender, line 151: I don't think our image is runnable on arm64 as I don't think we are building with arm as a target. Though I could be wrong about this.
```suggestion
          platforms: linux/amd64
```

jrhender: Hmm, maybe support for `arm64` will indeed work. We can give it a try.

.github/workflows/build-deploy.yml
jrhender: Maybe all we need to do is add another step that uploads the sentry images. 

We also need to pass the build arguments

```suggestion
      - name: Build and push Docker image
        uses: docker/build-push-action@v3
        with:
          context: .
          push: ${{  needs.configure.outputs.PUSH_DOCKER  }}
          platforms: linux/amd64,linux/arm64
          tags: ${{ steps.tags.outputs.TAGS }}
          labels: ${{ steps.meta.outputs.labels }}
          file: ./docker/prod/Dockerfile
          build-args: |
            SENTRY_RELEASE="${{ needs.configure.outputs.IMAGE_TAG }}"
          
     # Sentry upload is done via a docker build in order to have access to the app files in the image
     - name: create Sentry release
        uses: docker/build-push-action@v3
        with:
          context: .
          push: false
          file ./docker/sentry-upload/Dockerfile
          build-args: |
```

.github/workflows/build-deploy.yml
jrhender: good to know, thanks. I thinking you're saying that we can set `previous_tag` so that it will start from current version. I hope actually that won't be a problem as we do already have version tags.
I mean more that the version in the package.json file won't be updated: https://github.com/energywebfoundation/ssi-hub/blob/fff75267452ea4c411de560df5f837d0dc453c9e/package.json#L3
But I think that's okay, we can live with this mismatch for now and sort it out later.

jrhender (approved):
Looks good to me, let's give it a try

PR #582, opened 2023-02-17
task/SWTCH-2036 authenticate with siwe

JGiter:
Merge after https://github.com/energywebfoundation/passport-did-auth/pull/365


src/modules/auth/login.controller.ts
```diff
@@ -76,6 +83,47 @@ export class LoginController {
...
+      );
+    }
+
+    if ((await this.redis.get(nonce)) === 'false') {
+      throw new UnauthorizedException(
+        'Authentication with SIWE has completed already'
+      );
+    }
+
+    await this.redis.set(nonce, 'false');
```
jrhender, line 122: With this implementation, i think it would technically be possible that a nonce is reused. The sequence would be
1. a client submits a `verify` request to serverA and serverB
2. serverA gets the value and sees that it is not false
3. serverB gets the value and sees that it is not false
4. serverA sets the value to false
5. server B sets the value to false
6. both servers return the login response
I think that this isn't such a big problem but it could potentially lead to unexpected behaviour or security risk.
Perhaps a way to improve this would be to use a SET with the GET option (https://redis.io/commands/set/). Something like:
```suggestion
    if ((await this.redis.get(nonce)) === 'false') {
      throw new UnauthorizedException(
        'Authentication with SIWE has completed already'
      );
    }

    await this.redis.set(nonce, 'false');
    // if previousVal was true, then it would mean it was the first
    if (previousVal === 'true') {
       return await this.login(req, res);
    }
    else if (previousVal === 'false') {
      throw new UnauthorizedException(
        'Authentication with SIWE has completed already'
      );
   }
   else {
     // throw an error
   }
```

JGiter: You right. Thanks for noticing this case

jrhender:
Thanks for the update @JGiter . I noticed the e2e tests failed for the latest https://github.com/energywebfoundation/ssi-hub/pull/582/commits/afee347c973946ae2cb657a4d41f237344ea33a0 . Do you think it is an issue with the implementation or more test flakiness?

jrhender (approved):
Looks good to me, just pending the e2e test troubleshooting

jrhender:
@JGiter 
> Thanks for the update @JGiter . I noticed the e2e tests failed for the latest commit https://github.com/energywebfoundation/ssi-hub/commit/afee347c973946ae2cb657a4d41f237344ea33a0 (test run: https://github.com/energywebfoundation/ssi-hub/actions/runs/4240948219/jobs/7370581666). Do you think it is an issue with the implementation or more test flakiness?

I have triggered a re-execution of the tests: https://github.com/energywebfoundation/ssi-hub/actions/runs/4240948219/jobs/7371473423 👀

nichonien (approved):

src/modules/auth/login.controller.ts
artursudnik: the above means that if `this.login(req, res)` fails for some reason, another nonce needs to be generated. Correct?

nichonien:
> @jrhender Test is failing for two reasons now:
> 
> * we haven't merged corresponding PR In passport-did-auth yet
> * it turns out that `GET` modifier in redis command was added in redis 6.2 ([How to set AND get atomatically? Getting a sytax error redis/node-redis#1710 (comment)](https://github.com/redis/node-redis/issues/1710#issuecomment-961789484)), but we are currently using 6.0.8. Should we update redis?

@JGiter The PR in PDA is merged on and changes are available in version `2.0.0-alpha.27 `

JGiter:
@jrhender I am not sure should we also configure deployed Redis version? This PR passes probably because GA sets lates Redis when its version is not specified

src/modules/auth/login.controller.ts
```diff
@@ -76,6 +84,61 @@ export class LoginController {
...
+      });
+      // If authentication was initiated and not yet completed
+      if (isAuthenticating === 'true') {
+        return await this.login(req, res);
+      } else if (isAuthenticating === 'false') {
+        throw new UnauthorizedException(
+          'Authentication with SIWE has completed already'
+        );
+      } else {
+        throw new InternalServerErrorException(
```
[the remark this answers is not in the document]
JGiter: Indeed, will do this other way. Thanks

jrhender:
> should we also configure deployed Redis version?

@JGiter hmm, good question. I think we should indeed check if the Redis version that we have deployed will support the `set` with `get` option

artursudnik:
Strange, but I am having the following errors when executing `npm ci` locally. However, it executes correctly in CI. Tested on Linux.
```
$ npm ci
npm notice
npm notice New major version of npm available! 8.19.3 -> 9.5.0
npm notice Changelog: https://github.com/npm/cli/releases/tag/v9.5.0
npm notice Run npm install -g npm@9.5.0 to update!
npm notice
npm ERR! code ERESOLVE
npm ERR! ERESOLVE could not resolve
npm ERR!
npm ERR! While resolving: siwe@1.1.6
npm ERR! Found: ethers@5.7.0
npm ERR! node_modules/ethers
npm ERR!   ethers@"^5.7.0" from the root project
npm ERR!   ethers@"^5.0.13" from @ensdomains/ensjs@2.1.0
npm ERR!   node_modules/@ensdomains/ensjs
npm ERR!     @ensdomains/ensjs@"^2.0.1" from @nomiclabs/truffle-contract@4.5.10
npm ERR!     node_modules/@nomiclabs/truffle-contract
npm ERR!       @nomiclabs/truffle-contract@"^4.2.23" from @nomiclabs/hardhat-truffle5@2.0.6
npm ERR!       node_modules/@nomiclabs/hardhat-truffle5
npm ERR!         @nomiclabs/hardhat-truffle5@"^2.0.0" from @ensdomains/buffer@0.0.13
npm ERR!         node_modules/@ensdomains/buffer
npm ERR!   19 more (@ethereum-waffle/chai, @ethereum-waffle/compiler, ...)
npm ERR!
npm ERR! Could not resolve dependency:
npm ERR! peer ethers@"5.5.1" from siwe@1.1.6
npm ERR! node_modules/siwe
npm ERR!   siwe@"^1.1.6" from the root project
npm ERR!
npm ERR! Conflicting peer dependency: ethers@5.5.1
npm ERR! node_modules/ethers
npm ERR!   peer ethers@"5.5.1" from siwe@1.1.6
npm ERR!   node_modules/siwe
npm ERR!     siwe@"^1.1.6" from the root project
npm ERR!
npm ERR! Fix the upstream dependency conflict, or retry
npm ERR! this command with --force, or --legacy-peer-deps
npm ERR! to accept an incorrect (and potentially broken) dependency resolution.
npm ERR!
npm ERR! See /home/artur/.npm/eresolve-report.txt for a full report.

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/artur/.npm/_logs/2023-02-22T12_42_12_133Z-debug-0.log
```

jrhender:
@artursudnik 

> Strange, but I am having the following errors when executing npm ci locally. However, it executes correctly in CI. Tested on Linux.

See https://github.com/energywebfoundation/iam-client-lib/pull/678#discussion_r1108775739

artursudnik:
@jrhender 
> See https://github.com/energywebfoundation/iam-client-lib/pull/678#discussion_r1108775739

Looks like we need to update `npm ci` to `npm ci --legacy-peer-deps` in many places like Dockerfile and workflows.

.env.dev
```diff
@@ -11,7 +11,7 @@ DB_NAME=dev-test
 TYPEORM_LOGGING=error,migration,warn,info
 
 # REDIS
-REDIS_VERSION=6.0.8
+REDIS_VERSION=7
```
jrhender, line 14: I would go back to the older version here as this better matches what we have in the deployed infra
```suggestion
REDIS_VERSION=6.0.8
```

jrhender (commented):
@JGiter related to @artursudnik observations, maybe we should add `--legacy-peer-deps` in all the locations with `npm ci` as a part of this PR. What do you think?

JGiter:
> @JGiter related to @artursudnik observations, maybe we should add `--legacy-peer-deps` in all the locations with `npm ci` as a part of this PR. What do you think?

Yes, I guess we need this, though I don't have errors with `npm i` or `npm ci`. Also instead of multiple changes, may be we should add `.npmrc` and put `--legacy-peer-deps` there @jrhender ?

jrhender:
@JGiter 
> Also instead of multiple changes, may be we should add .npmrc and put --legacy-peer-deps there @jrhender ?

Yes, I'm open to this. Though will the .npmrc file get pickup by GitHub actions jobs? We might need to add `--legacy-peer-deps` directly in the workflow files

artursudnik:
> Though will the .npmrc file get pickup by GitHub actions jobs?

I'd be really surprised if not.

jrhender:
> > Though will the .npmrc file get pickup by GitHub actions jobs?
> 
> I'd be really surprised if not.

Ah yes, makes sense. I wasn't think it through. Of course GHA will have access to the `.npmrc` as it will be checking out the repo before executing `npm ci` 👍

PR #586, opened 2023-02-20
task/update passport-did-auth@@2.0.0-alpha.26

JGiter:
We need to update passport-did-auth to the version used in development in order to integrate recent changes of SIWE authentication


package.json
```diff
@@ -53,6 +53,7 @@
     "@ensdomains/resolver": "^0.2.4",
     "@ew-did-registry/claims": "0.7.1-alpha.795.0",
     "@ew-did-registry/credentials-interface": "0.7.1-alpha.795.0",
+    "@ew-did-registry/did": "0.7.1-alpha.795.0",
```
JGiter, line 56: Since this direct dependency it should be saved in package.json. Without it can not link `passport-did-auth` locally, because `npm link` installs dependency of `passport-did-auth` in `passport-did-auth` directory and thus makes `@ew-did-registry/did` unavailable for `ssi-hub`

.github/workflows/code-lint.yml
```diff
@@ -7,6 +7,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+      - name: Setup Node.js
```
JGiter, line 10: Without this Node.js 18 is used, which leads to https://github.com/energywebfoundation/ssi-hub/actions/runs/4222857695/jobs/7331874584

src/modules/claim/claim-verification.service.spec.ts
```diff
@@ -27,7 +27,7 @@ const resolvedCredential: RoleEIP191JWT = {
...
   payload: {
     claimData: {
-      fields: undefined,
+      requestorFields: undefined,
```
JGiter, line 30: Apparently updating of `passport-did-auth` caused updating of `@energyweb/vc-verification`, which allowed to detect this error

src/modules/auth/login.strategy.ts
```diff
@@ -2,29 +2,41 @@ import { LoginStrategy } from 'passport-did-auth';
...
+      issuerResolver,
+      revokerResolver,
+      credentialResolver,
+      verifyCredential,
+    ];
+    const numberOfBlocksBack = parseInt(
+      configService.get<string>('STRATEGY_NUM_BLOCKS_BACK')
+    );
+    if (numberOfBlocksBack) {
+      loginStrategyParams[0].numberOfBlocksBack = numberOfBlocksBack;
```
jrhender, line 38: I think it's slightly nicer to keep a reference to `loginStrategyOptions` so that `numberOfBlocksBack` can be added without using `0` index. But not a big deal.
```suggestion
    const loginStrategyOptions: LoginStrategyOptions = {
        name: 'login',
        rpcUrl: configService.get<string>('ENS_URL'),
        cacheServerUrl: configService.get<string>('STRATEGY_CACHE_SERVER'),
        privateKey: configService.get<string>('STRATEGY_PRIVATE_KEY'),
        didContractAddress: configService.get<string>('DID_REGISTRY_ADDRESS'),
        ensRegistryAddress: configService.get<string>('ENS_REGISTRY_ADDRESS'),
        ipfsUrl: ipfsConfig,
     }
    const loginStrategyParams: ConstructorParameters<typeof LoginStrategy> = [
      loginStrategyOptions,
      issuerResolver,
      revokerResolver,
      credentialResolver,
      verifyCredential,
    ];
    const numberOfBlocksBack = parseInt(
      configService.get<string>('STRATEGY_NUM_BLOCKS_BACK')
    );
    if (numberOfBlocksBack) {
      loginStrategyOptions.numberOfBlocksBack = numberOfBlocksBack;
```

It should work, I did a quick test in node repl:
```
> const nested = {}
undefined
> const outer = [nested]
undefined
> nested.prop = 1
1
> outer
[ { prop: 1 } ]
```

jrhender (approved):
Looks good to me 👍

src/modules/auth/login.strategy.ts
JGiter: yes, it allowes accidentialy assign `numberOfBlocksBack` to some other property of loginStrategyOptions

PR #591, opened 2023-02-22
Remove code security warning about 'HttpOnly' flag

jrhender:
I think the behaviour of the security scan is such that if the variable could change between declaration and usage, it flags it as an issue.

Relates to: https://energyweb.atlassian.net/browse/SWTCH-1888

Note: also contains documentation updates from https://github.com/energywebfoundation/ssi-hub/pull/596


artursudnik (approved):

nichonien:
@jrhender These changes were also suggested by `Snyk` and it resolves the issues. But IMO this forces developer to provide cookiesSettings explicitly for every response. It should be possible (don't know the exact way) to use custom / same cookie settings for all (avoiding redundancy).

JGiter (approved):

src/modules/auth/login.controller.ts
```diff
@@ -20,16 +20,19 @@ import { SiweMessage, generateNonce } from 'siwe';
...
 import { RoleService } from '../role/role.service';
 import { SiweReqPayload } from './siwe.dto';
 
 @ApiTags('Auth')
 @Controller({ version: '1' })
 export class LoginController {
+  // Using "private readonly" class members so that automated scan can be certain that cookie options are set appropriately
+  private readonly cookieIsHttpOnly = true;
+  private readonly cookieIsSecure = true;
+  private readonly cookieSameSite = 'none';
```
jrhender, line 32: @nichonien how about this approach? Does this strike a better balance between reuse and immutability?

jrhender:
Docs changes shouldn't be in this PR. I will rebase after https://github.com/energywebfoundation/ssi-hub/pull/597

src/modules/auth/login.controller.ts
nichonien: Looks good for now, we can merge this change. It would be good to understand (some time later) why the scan doesn't pick the `cookieOptions` being imported or how exactly the code scan works.

nichonien (approved):

nichonien (approved):

JGiter (approved):

PR #592, opened 2023-02-22
chore: upgrading node to v18 in all GH actions

artursudnik:


artursudnik:
closing this because `@poanet/solidity-flattener` requires v16

PR #594, opened 2023-02-22
fix: fixing lockfile issue

artursudnik:
This PR:
- prevents unintentional modification of the package-lock.json by npm v8
- downgrades docker image base image to ensure it uses npm v7 to execute `npm ci`


docker/prod/Dockerfile
```diff
@@ -1,4 +1,4 @@
-ARG IMAGE=node:18-alpine
+ARG IMAGE=node:16.10.0-alpine
```
artursudnik, line 1: However, I am not sure if we should keep v18 and install npm v7.

jrhender:
@artursudnik can you take a look at the `.npmrc` conflict when you have a moment please?

docker/prod/Dockerfile
jrhender: I'm leaning towards keeping 16.10 as this is what we have in the tests: https://github.com/energywebfoundation/ssi-hub/blob/744754e73dfc9f2b835b4202a3e910602ca566f0/.github/workflows/build-test.yml#L67

jrhender (approved):

JGiter (approved):

PR #596, opened 2023-02-24
bug/type of siwe message dto

JGiter:


jrhender (approved):

nichonien (approved):

jrhender:
@JGiter if you think it makes sense, I think it could be good to add tests that cover this case

artursudnik (approved):

PR #599, opened 2023-02-28
task/SWTCH-2101 option to disable block number authentication

JGiter:


.github/workflows/build-test.yml
```diff
@@ -50,6 +50,7 @@ env:
   NESTJS_PORT: ${{secrets.NESTJS_PORT}}
   STRATEGY_CACHE_SERVER: ${{secrets.STRATEGY_CACHE_SERVER}}
   STRATEGY_PRIVATE_KEY: ${{secrets.STRATEGY_PRIVATE_KEY}}
+  BLOCKNUM_AUTH_ENABLED: true
```
JGiter, line 53: this is temporarily until we change e2e tests to authenticate with Siwe

devops/prod/values.yaml
```diff
@@ -111,6 +111,7 @@ iam-cache-server-helm:
     JWT_PRIVATE_KEY_PATH: /keys/private.pem
     JWT_PUBLIC_KEY_PATH: /keys/public.pem
     STRATEGY_CACHE_SERVER: https://identitycache.energyweb.org/v1
+    BLOCKNUM_AUTH_ENABLED: false
```
jrhender, line 114: Should be enabled in prod to not break existing apps. We can disable once we've had time to migrate all of the clients.
```suggestion
    BLOCKNUM_AUTH_ENABLED: true
```

devops/staging/values.yaml
```diff
@@ -114,6 +114,7 @@ iam-cache-server-helm:
     JWT_PRIVATE_KEY_PATH: /keys/private.pem
     JWT_PUBLIC_KEY_PATH: /keys/public.pem
     STRATEGY_CACHE_SERVER: https://identitycache-staging.energyweb.org/v1
+    BLOCKNUM_AUTH_ENABLED: false
```
jrhender, line 117:
```suggestion
    BLOCKNUM_AUTH_ENABLED: true
```

src/modules/auth/auth.module.ts
```diff
@@ -68,6 +71,16 @@ export class AuthModule implements NestModule {
...
+      .apply((_req: Request, res: Response) => {
+        if (!this.configService.get('BLOCKNUM_AUTH_ENABLED')) {
+          res
+            .status(404)
+            .send(
+              'Authentication at this endpoint is disabled. Other authentication protocols may be available'
+            );
+        }
+      })
+      .forRoutes({ path: '/v1/login', method: RequestMethod.POST });
```
jrhender, line 84: Why have the `404` logic as a middleware?
Alternatively, the logic could be in the `Controller`. Maybe middleware is better, I'm not sure. For instance, maybe middleware is faster. Just curious what your perspective is.

jrhender (changes requested):
The config for dev environment should have Blocknumber enabled

src/modules/auth/auth.module.ts
artursudnik: maybe NestJS has some decorator where this logic could be implemented? Like guards. I am not sure...

it would be the cleanest approach.

JGiter: From my opinion `/login` hanldler is responsible for configuring session after succesful login. If endpoint should not be accessible, then this configuring and authentication is not needed

artursudnik: yeah, middleware is executed very early, co nothing more is initiated if it is disabled like this

JGiter: Probably interceptor https://docs.nestjs.com/interceptors is a more appropriate option for this

JGiter: intercetpor could be a more declarative approach, but it is after Guard

jrhender (commented):
The PR looks good but we shouldn't merge until we've fixed the ability of passport-did-auth to authenticate to an ssi-hub that only supports SIWE (e.g. supporting SIWE in `cacheServerClient` https://github.com/energywebfoundation/passport-did-auth/blob/develop/lib/cacheServerClient.ts)

jrhender (approved):

src/modules/auth/auth.module.ts
```diff
@@ -68,6 +71,17 @@ export class AuthModule implements NestModule {
...
+      .forRoutes({ path: '/*', method: RequestMethod.ALL })
+      .apply((_req: Request, res: Response, next: NextFunction) => {
+        if (!this.configService.get('BLOCKNUM_AUTH_ENABLED')) {
+          res
+            .status(404)
+            .send(
+              'Authentication at this endpoint is disabled. Other authentication protocols may be available'
+            );
+        }
+        next();
```
jrhender, line 83: @JGiter when testing against a ssi-hub running locally on my machine, I'm seeing a possible issue

Repro steps:
1. Run ssi-hub locally with `BLOCKNUM_AUTH_ENABLED=false`
2. Try to authenticate against `/login` endpoint (e.g. with ICL < v8)

Observed result:
Following error is logged in ssi-hub
```
[Nest] 90161  - 03/01/2023, 3:15:22 PM   ERROR [ExceptionsHandler] Cannot set headers after they are sent to the client
Error: Cannot set headers after they are sent to the client
    at new NodeError (node:internal/errors:371:5)
    at ServerResponse.setHeader (node:_http_outgoing:576:11)
    at ServerResponse.header (/Users/john/ew/ssi-hub/node_modules/express/lib/response.js:794:10)
    at ServerResponse.append (/Users/john/ew/ssi-hub/node_modules/express/lib/response.js:755:15)
    at ServerResponse.res.cookie (/Users/john/ew/ssi-hub/node_modules/express/lib/response.js:884:8)
    at LoginController.login (/modules/auth/login.controller.ts:73:9)
```
To me, the above suggests that maybe `login` method of `login.controller` is still being executed after response is being sent.

Maybe the issue is with this `next()` call

jrhender (changes requested):

src/modules/auth/auth.module.ts
jrhender: Yes, after removing `next`, error is not logged in ssi-hub
```suggestion
```

JGiter: may be `next()` is not needed at all. I'll check this issue. thanks to catching this

JGiter: the issue was not caught becasue there is no straight way to change environment variables from test. We need to figure out how this can be improved

src/modules/auth/auth.module.ts
```diff
@@ -68,6 +71,16 @@ export class AuthModule implements NestModule {
         { path: '/v1/login/siwe/initiate', method: RequestMethod.POST },
         { path: '/v1/login/siwe/verify', method: RequestMethod.POST }
       )
-      .forRoutes({ path: '/*', method: RequestMethod.ALL });
+      .forRoutes({ path: '/*', method: RequestMethod.ALL })
```
jrhender, line 74: Would it be better just to have `POST` here as it is only the `POST` endpoint that we are trying to disable (not that there is a `GET` at this point)?
```suggestion
      .forRoutes({ path: '/*', method: RequestMethod.POST })
```

src/modules/auth/auth.module.ts
```diff
@@ -68,6 +71,16 @@ export class AuthModule implements NestModule {
...
+      .apply((_req: Request, res: Response, next: NextFunction) => {
```
jrhender, line 75: `next` can be removed to fix warning
```suggestion
      .apply((_req: Request, res: Response) => {
```

src/modules/auth/auth.module.ts
```diff
@@ -68,6 +71,16 @@ export class AuthModule implements NestModule {
...
+        if (!this.configService.get('BLOCKNUM_AUTH_ENABLED')) {
+          res
+            .status(404)
+            .send(
+              'Authentication at this endpoint is disabled. Other authentication protocols may be available'
```
jrhender, line 80: I think returning an object as a response would be more ideomatic
This would better match other 404 responses. See, for example, this attempt to `GET` to `login` endpoint:
```
❯ curl -X GET https://identitycache.energyweb.org/v1/login          
{"statusCode":404,"message":"Cannot GET /v1/login","error":"Not Found"}
```
```suggestion
{
    code: 404,
    error: 'Not Found',
    message: 'Authentication at this endpoint is disabled. Other authentication protocols may be available'
}    
```

jrhender (commented):
Some other small minor improvements/questions

src/modules/auth/auth.module.ts
```diff
@@ -68,6 +71,16 @@ export class AuthModule implements NestModule {
...
+            );
+        }
+      })
+      .forRoutes({ path: '/v1/login', method: RequestMethod.ALL });
```
jrhender, line 84: Would it be better to just have `POST` here as it is the only endpoint that we're trying to disable? Better to be precise, I think. For example, if we ever did implement a `GET` endpoint, it could be strange behaviour.
```suggestion
      .forRoutes({ path: '/v1/login', method: RequestMethod.POST });
```

src/modules/auth/auth.module.ts
JGiter: In my opinion status code and message serves different pusposes, so they should not be mixed. But I guess it will not hurt in include status in message
By the way you can get status by curl with `-i` option https://curl.se/docs/manpage.html#-i

jrhender (approved):

PR #600, opened 2023-03-02
Cancel previous build-test workflow

JGiter:


jrhender (commented):
Was this just an omission from the workflow previously and you noticed that it was need @JGiter ?

JGiter:
> Was this just an omission from the workflow previously and you noticed that it was need @JGiter ?

Yes, it was not added initially and it leads to multipe unnecessary builds like 859, 860 and 861 https://github.com/energywebfoundation/ssi-hub/actions/workflows/build-test.yml

PR #602, opened 2023-03-11
bug/SWTCH-2177 authentication status does not respect origin

JGiter:
Authentication status will consider origin of access token


src/modules/auth/token.service.spec.ts
```diff
@@ -106,18 +106,14 @@ describe('TokenService', () => {
...
+  it('should authorize if token origin matches request origin', async () => {
     return mockRequest('get', 'https://switchboard-dev.energyweb.org');
   });
 
-  it('should NOT have same POST request ORIGIN header as stored in JWT token ', async () => {
+  it('should not authorize if token origin does not matches request origin ', async () => {
     return mockRequest('post', 'https://hacked-website.com', 401);
   });
 
-  it('should NOT have origin in request headers', async () => {
```
JGiter, line 117: removed because this test allows empty request origin when origin of access token is set

nichonien (approved):

src/modules/auth/token.service.ts
```diff
@@ -114,18 +118,14 @@ export class TokenService {
...
     const isServerRequestOrGETFromSameDomain =
       req.headers['origin'] === undefined;
-    if (
-      !isBrowserRequestFromAuthenticatedOrigin &&
-      !isServerRequestOrGETFromSameDomain
-    ) {
-      throw new UnauthorizedException(
-        `Token origin ${origin} does not matches request origin ${req.headers['origin']}`
-      );
-    }
```
jrhender, line 129: Why the removal of the exception throwing?

JGiter: This is because we are not always throw Unauthorized, if login origins are not matches. `/auth/status` in this case still returns status 200 just with `null` user. So I changed `matchOriginAgainstRequest` to `isLoginOriginMatchesRequestOrigin` to be responsible only for matching and letting client decide how to respond to result of matching

e2e/auth/authStatusBlockNum.testSuite.ts
```diff
@@ -0,0 +1,170 @@
...
+        });
+
+        describe('when request origin is not set', () => {
+          it.todo(
+            'should return negative logged in status for cookies session'
+          );
+
+          it.todo(
+            'should return negative logged in status for authorization header'
+          );
```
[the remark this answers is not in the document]
JGiter: This tests are intended to check that request origin matches login origin, but our current implementation does not require this. When request origin is empty, then login origin is not considered https://github.com/energywebfoundation/ssi-hub/blob/504989287e9e257ebc9518dd060d822d4ec026e9/src/modules/auth/token.service.ts#L120. So I added this todo's just as a reminder

src/modules/auth/login.controller.ts
```diff
@@ -218,6 +227,15 @@ export class LoginController {
         accessTokenString
       );
 
+      if (
+        !this.tokenService.isLoginOriginMatchesRequestOrigin(
+          tokenData?.origin,
+          req
+        )
+      ) {
+        throw new Error();
```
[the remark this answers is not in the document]
JGiter: Actually nobody will see this error. Throwing here is just a way to return `user: null`, because this error will be caught and processed later. It is possible to achieve same without throwing, but I choose this way to be consistent with case when refresh token not verified (line 169)

src/modules/auth/login.controller.ts
```diff
@@ -154,7 +154,16 @@ export class LoginController {
       throw new UnauthorizedException('Refresh token was not set');
     }
 
-    this.tokenService.matchOriginAgainstRequest(refreshTokenString, req);
+    if (
+      !this.tokenService.isLoginOriginMatchesRequestOrigin(
+        refreshTokenString,
```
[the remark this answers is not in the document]
JGiter: Right, we have task to fix this in another PR

e2e/auth/login.testSuite.ts
```diff
@@ -218,99 +218,141 @@ export const authLoginTestSuite = () => {
...
+      describe('when SIWE message is invalid', () => {
+        beforeAll(async () => {
+          const { text } = await request(app.getHttpServer())
+            .post('/v1/login/siwe/initiate')
+            .expect(201);
+          const nonce = JSON.parse(text).nonce;
+          ({ message, signature } = await signSiweMessage(wallet, nonce));
+        });
+
+        it('should not verify unprepared message', async () => {
```
artursudnik, line 346: maybe `it should respond with 400 status code` would be more precise here?

e2e/auth/login.testSuite.ts
```diff
@@ -218,99 +218,141 @@ export const authLoginTestSuite = () => {
...
+              .get('/v1/search/test')
+              .set('Cookie', [
+                loginResponse.headers['set-cookie'][0].split(';')[0] + ';',
+                loginResponse.headers['set-cookie'][1].split(';')[0] + ';',
+              ])
+              .set('Origin', origin)
+              .expect(200);
+          });
+
+          it('should not be authorized when request origin does not matches origin of access token', async () => {
```
artursudnik, line 266: maybe specifying what should technically happen would be better? This description is ambiguous.

e2e/auth/utils.ts
```diff
@@ -0,0 +1,58 @@
...
+    .expect(201);
+  const nonce = JSON.parse(text).nonce;
+  const { message, signature } = await signSiweMessage(wallet, nonce);
+  const req = request(app.getHttpServer()).post('/v1/login/siwe/verify').send({
+    message,
+    signature,
+  });
+
+  if (origin) {
+    req.set('Origin', origin);
```
artursudnik, line 53: why is this set here after the request has been already sent?

e2e/auth/login.testSuite.ts
JGiter: Yes, I am personally prefer terse descriptions. Full version would be 
`Authorized endpoint should return Unauthorized exception, when request origin does not match origin of acees token`

Does this sounds more precise from your perspective?

e2e/auth/utils.ts
JGiter: I don't think request is sent before await. In editor I can see that type of `req` is `supertest.Test`, but type of `loginResponse` is `supertest.Response`

e2e/auth/login.testSuite.ts
artursudnik: why not describe what actually is checked? In this case response status code, not exception. It really helps when something fails in tests and it is possible to see what was wrong without looking into the test.

e2e/auth/utils.ts
artursudnik: you are right. nevermind :)

e2e/auth/login.testSuite.ts
artursudnik: what about: `when request origin does not match access token origin, it should respond with `401` status code?
this is what actually is tested, not the exception returned.

JGiter: Ok, sounds good

jrhender (approved):
Looks good to me, thanks @JGiter

PR #612, opened 2023-03-21
chore: add logs for domain update events

whitneypurdum:


jrhender (changes requested):
We don't need to merge into master branch to deploy to gp4btc-staging and there is still a code freeze in staging env as far as I know. I think this PR should be closed.

PR #614, opened 2023-03-22
Log role dto

whitneypurdum:
To debug issue of data fetch not returning the JSON schema:
- log out credential issuer field data before post to db
- log dto data before post

### DIFF
--- a/src/modules/role/role.service.ts
+++ b/src/modules/role/role.service.ts
@@ -365,7 +365,7 @@ export class RoleService {
       return;
     }
     this.logger.debug(
-      `### Updating DTO for namespace ${namespace}, DTO: ${JSON.stringify(dto)}`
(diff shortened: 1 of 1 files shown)


src/modules/role/role.service.ts
```diff
@@ -360,7 +364,9 @@ export class RoleService {
       );
       return;
     }
-    this.logger.debug(`### Updating DTO for namespace ${namespace}`);
+    this.logger.debug(
+      `### Updating DTO for namespace ${namespace}, DTO: ${JSON.stringify(dto)}`
+    );
```
jrhender, line 369: In my opinion, the `###` are noise in the logs
```suggestion
    this.logger.debug(
      `Updating DTO for namespace ${namespace}, DTO: ${JSON.stringify(dto)}`
    );
```

jrhender (approved):
Looks good, just a suggestion re "###"

PR #623, opened 2023-03-30
fix(CD): fixing non-parseable `build.json` file

artursudnik:


.github/workflows/build-deploy.yml
```diff
@@ -145,7 +145,7 @@ jobs:
           file: ./docker/prod/Dockerfile
           build-args: |
             SENTRY_RELEASE="${{ needs.configure.outputs.IMAGE_TAG }}"
-            GIT_SHA="${{ github.sha }}}"
+            GIT_SHA=${{ github.sha }}}
```
JGiter, line 148: just curious, why this expression should not be quoted? Also there is extra closed `}`
```suggestion
            GIT_SHA=${{ github.sha }}
```

JGiter (changes requested):

.github/workflows/build-deploy.yml
artursudnik: oh! maybe that extra `}` is the problem! thanks for noticing, I am adding back quotation marks and fixing this.

JGiter (approved):

PR #628, opened 2023-04-06
fix: remove ens-sync on app start

nichonien:


src/modules/ens/ens.service.ts
```diff
@@ -124,7 +124,6 @@ export class EnsService implements OnModuleDestroy {
       );
       this.schedulerRegistry.addInterval('ENS Sync', interval);
       this.InitEventListeners();
-      this.syncENS();
```
jrhender, line 127: Rather than making a code change here, I think we should try to change at configuration level so that it isn't a hotfix. We could set `ENS_SYNC_ENABLED = false` instead

jrhender (changes requested):

jrhender (approved):
Actually, maybe this is the better solution so that the container can start and serve requests for some time but we still get the benefit of ENS sync

PR #631, opened 2023-04-24
ens sync update

nichonien:


JGiter (approved):
@nichonien This PR awaits for some time. In my opinion this addition makes to think that RoleDefinitionV2 is not RoleDefiniton, but I am also not against merging it

PR #633, opened 2023-04-28
task/SWTCH-2208 role not updated

JGiter:
Just a couple logs to track failed role updates

### DIFF
--- a/src/modules/role/role.service.ts
+++ b/src/modules/role/role.service.ts
@@ -139,8 +139,14 @@ export class RoleService {
     if (!role) return this.create(data);
 
     if (data.appNamespace) {
+      this.logger.debug(
+        `Updating role ${data.name} defined under ${data.appNamespace} application`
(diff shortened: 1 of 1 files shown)


src/modules/role/role.service.ts
```diff
@@ -139,8 +139,14 @@ export class RoleService {
...
 
     if (data.appNamespace) {
+      this.logger.debug(
+        `Updating role ${data.name} defined under ${data.appNamespace} application`
+      );
       const app = await this.appService.getByNamespace(data.appNamespace);
       if (!app) {
+        this.logger.warn(
+          `Can not update role ${data.name}: parent application ${data.appNamespace} is not cached`
+        );
```
jrhender, line 149: Maybe add an equivalent log to the `if (!org)` branch as well

jrhender (approved):
Seems okay to me 👍

PR #636, opened 2023-05-12
task/SWTCH-2331 disable get dids by role

JGiter:


jrhender (commented):
looks like a good start to me 👍

jrhender (commented):
looks like a good start to me 👍

nichonien (approved):

PR #639, opened 2023-05-15
fix: move cjs-exporter/p-map to dep

nichonien:


JGiter (commented):
@nichonien why ssi-hub needs @cjs-exporter/p-map? I thought only ew-credential/credential-governance uses it

nichonien:
> @nichonien why ssi-hub needs @cjs-exporter/p-map? I thought only ew-credential/credential-governance uses it

Yeah, I don't know why this is a requirement. Maybe because it is not explicitly defined in `ew-credential/credential-governance` package-json, but in the main?

PR #647, opened 2023-05-31
Publish v2.12.1

JGiter:


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+### [2.12.1](https://github.com/energywebfoundation/ssi-hub/compare/v2.12.0...v2.12.1) (2023-05-31)
```
jrhender, line 5: Strange that there are no changes in the changelog. Is this expected?

JGiter: There were no changes in develop, although `standard-version` bumped version. I merged one fix, so now there will be actual change

jrhender: Ah okay, thanks for clarifying. I thought that maybe the disabling of claims/{did} might show up as a change but perhaps not

jrhender (approved):
Looks good to me. A note that this should be merged with a "merge commit" and not a squash so that the tagged commit is maintained.

CHANGELOG.md
JGiter: claims/did/role was disabled in previous version v2.12.0. It is not shown in CHANGELOG.md, because `conventional-commits` preset used by standard-release by default hides `refactor` type commit https://github.com/conventional-changelog/conventional-changelog-config-spec/blob/master/versions/2.2.0/README.md#types

PR #649, opened 2023-06-09
Purge extraneous role claims

JGiter:


src/scripts/purge-extraneous-claims.ts
```diff
@@ -16,20 +17,33 @@ import { RoleClaim } from '../modules/claim/entities/roleClaim.entity';
...
   const claims = await claimsRepository.find();
   for (const claim of claims) {
     if (
       claim.vp.verifiableCredential.some(
         (vc) =>
+          vc.credentialStatus &&
           !vc.credentialStatus.statusListCredential.startsWith(
             STATUS_LIST_DOMAIN
           )
       )
```
jrhender, line 30: For this case of cleaning non-gp4btc data, I think that there may be `claims` that we want to remove that don't have a VP or don't have VP with a VC with `credentialStatus`. 

Does this logic make sense to accomplish this? My idea is that if the credential doesn't have a VC with the `STATUS_LIST_DOMAIN` that we expect, then we can remove it.
```suggestion
      !claim?.vp?.verifiableCredential.some(
        (vc) =>
          vc.credentialStatus &&
          vc.credentialStatus.statusListCredential.startsWith(
            STATUS_LIST_DOMAIN
          )
      )
```

JGiter: Yes, this sounds right

jrhender (approved):

PR #652, opened 2023-06-22
SWTCH-2439 claim not synced

JGiter:
Trying to handle two issues:
- stalled pinning to cluster blocks synchronization of document, because pinning is part of synchronization job. Now jobs are separate
- stalled synchronization job block other jobs. ~~Now jobs will be ran in separate processes~~. Unfortunatelly this forces to give up nestjs injection https://docs.nestjs.com/techniques/queues#separate-processes. Instead pinning jobs are added in another queue. This way pinning will not block synchronization, but jobs in the same queue can still block each other

Also refactored debug logs from handlers

Locally tested that this is fixes publishing issue

### DIFF
--- a/src/modules/did/did.processor.ts
+++ b/src/modules/did/did.processor.ts
@@ -1,9 +1,9 @@
 import {
   InjectQueue,
-  OnGlobalQueueActive,
-  OnGlobalQueueError,
-  OnGlobalQueueFailed,
-  OnGlobalQueueStalled,
+  OnQueueActive,
+  OnQueueError,
+  OnQueueFailed,
+  OnQueueStalled,
   Process,
   Processor,
 } from '@nestjs/bull';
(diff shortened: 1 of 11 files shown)


src/modules/did/did.processor.ts
```diff
@@ -1,92 +1,78 @@
...
-      if (this.configService.get<boolean>('DID_SYNC_MODE_FULL')) {
-        doc = await this.didService.addCachedDocument(job.data, true);
-      } else {
-        doc = await this.didService.incrementalRefreshCachedDocument(job.data);
-      }
+  @OnGlobalQueueFailed()
+  async onFailed(jobId: number) {
+    const job = await this.queue.getJob(jobId);
+    this.logger.debug(`Failed ${job.name} document ${job.data}`);
+  }
```
jrhender, line 58: @JGiter why use listeners for Global events, rather than local events?
I found this documentation discussing this but it's not clear to me why we would want Global listeners in our case https://github.com/OptimalBits/bull/blob/develop/REFERENCE.md#global-events
May Global is better than Local, it's just not clear to me why this would be.

jrhender:
@JGiter In the PR description you wrote:
> ~~Now jobs will be ran in separate processes~~. Unfortunatelly this forces to give up nestjs injection https://docs.nestjs.com/techniques/queues#separate-processes.

Should the "Unfortunatelly this forces to give up nestjs injection https://docs.nestjs.com/techniques/queues#separate-processes." bit be crossed out as well? It looks like we are still using NestJs DI everywhere

src/modules/did/did.processor.ts
JGiter: I registered listeners globaly becasue I am uncertaing about Redis configuration and afraid that events can be lost somehow. Perhaps it doesn't really make sence. I'll set listeners locally

JGiter:
> Should the "Unfortunatelly this forces to give up nestjs injection https://docs.nestjs.com/techniques/queues#separate-processes." bit be crossed out as well? It looks like we are still using NestJs DI everywhere

Sorry, this description is confusing. Giving up injection relates only to jobs in separate processes. It explains why we didn't follow that approach

jrhender (approved):
Looks good, thanks @JGiter

PR #657, opened 2023-07-05
SWTCH-2330 Remove claims by subjects endpoint

JGiter:


nichonien (approved):

jrhender (approved):
Looks good, thanks @JGiter . I think we could merge however a consideration could be to wait until we have solved the current issues in the dev environment, in order to not complicate the change history of the branch while we are investigating. If you think it doesn't affect the investigation, feel free to merge 👍

JGiter:
> I think we could merge however a consideration could be to wait until we have solved the current issues in the dev environment, in order to not complicate the change history of the branch while we are investigating

It is unlikely code issue. Besides the issues are no longer occures

PR #659, opened 2023-07-26
Claims of updated document are not pinned

JGiter:
This fixes recently occurred `Error pinning claims Unexpected token o in JSON at position 1` bugs

### DIFF
--- a/src/modules/did/did.processor.ts
+++ b/src/modules/did/did.processor.ts
@@ -69,6 +69,10 @@ export class DIDProcessor {
       doc = await this.didService.incrementalRefreshCachedDocument(job.data);
     }
 
-    await this.pinQueue.add(PIN_CLAIM_JOB_NAME, doc);
+    for (const service of doc.service) {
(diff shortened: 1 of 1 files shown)


jrhender:
@JGiter could you add a bit of description to the PR please? Is this a bug fix?

src/modules/did/did.processor.ts
```diff
@@ -69,6 +69,10 @@ export class DIDProcessor {
       doc = await this.didService.incrementalRefreshCachedDocument(job.data);
     }
 
-    await this.pinQueue.add(PIN_CLAIM_JOB_NAME, doc);
+    for (const service of doc.service) {
+      await this.pinQueue.add(PIN_CLAIM_JOB_NAME, {
+        cid: service.serviceEndpoint,
+      });
+    }
```
jrhender, line 76: I would say that it's not important to change here, but curious if there is a reason that you choose a `for` loop over `map` & `Promise.all`?

JGiter: No reason. I just can't get used to Promise.all. I changed that

PR #660, opened 2023-08-03
Exceeded DID queue memory limit

JGiter:
Handles errors on adding pin and DID synchronization jobs

### DIFF
--- a/docs/api/classes/modules_did_did_processor.DIDProcessor.md
+++ b/docs/api/classes/modules_did_did_processor.DIDProcessor.md
@@ -10,6 +10,7 @@
 
 ### Methods
 
+- [OnQueueWaiting](modules_did_did_processor.DIDProcessor.md#onqueuewaiting)
 - [onActive](modules_did_did_processor.DIDProcessor.md#onactive)
 - [onError](modules_did_did_processor.DIDProcessor.md#onerror)
 - [onFailed](modules_did_did_processor.DIDProcessor.md#onfailed)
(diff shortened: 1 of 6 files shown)


jrhender (commented):
I think this PR is okay as a fix, but not ideal.

One edge case that I'm worried that this could introduce is a scenario like this:
1. A user calls the IPFS service to pin some credentials
2. The user updates a DID document with services that reference the IPFS hashes
3. The pin from step 1 fails, but the DID document update from step 2 succeeds. Now the user would have a DID Document whose service endpoints are not accessible.

What do you think of this edge case @JGiter ?

I think we should consider splitting this PR into:
1. A PR which adds the new logs (which I think it can't hurt to have)
2. A PR which adds the `queue.empty()` calls, as I think these are a bit dangerous from a data-loss perspective

What are your thoughts @JGiter ?

JGiter:
@jrhender I agree that queue purging is not most graceful way and we can skip it for now. From the other side jobs removal does not affect outcome of adding credentials to IPFS service or getting of document. Call to save in IPFS service fails only when data not persisted neither in Infura nor in cluster, but in this case IPFS service will not return CID to include in document https://github.com/energywebfoundation/ssi-hub/blob/291ab134d182a9fb658fbe56d469172f23840711/src/modules/ipfs/ipfs.service.ts#L78. Resolution of document doesn't depend on whether claims were saved in cluster. If claim is not in cluster it will be resolved from Infura.

jrhender (approved):
Looks good to me, thanks @JGiter

PR #665, opened 2023-09-19
Feat/swtch 2493/no master build

jrhender:
This PR:
- Uses released ssi-hub image for deploy to staging (i.e. `master` branch)
- Removes build of new image when merging to `master`. We can put this back later if staging env is disconnected from `master` branch in the future

@Timtech4u adding you as reviewer as you helped setup the ssi-hub build for gp4btc so you might remember something about this area

### DIFF
--- a/devops/staging/values.yaml
+++ b/devops/staging/values.yaml
@@ -7,7 +7,7 @@ iam-cache-server-helm:
 
   image:
     repository: "ghcr.io/energywebfoundation/ssi-hub"
-    tag: dce9b11b-2242-4b17-8087-e8bfbcb98d98
+    tag: v2.13.0
     pullPolicy: IfNotPresent
 
   imagePullSecrets: [ ]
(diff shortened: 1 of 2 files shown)


devops/staging/values.yaml
```diff
@@ -7,6 +7,7 @@ iam-cache-server-helm:
 
   image:
     repository: "ghcr.io/energywebfoundation/ssi-hub"
+    tag: dce9b11b-2242-4b17-8087-e8bfbcb98d98
```
jrhender, line 10: I want to merge `develop` to `release` to get a new stable release tomorrow and then use this release instead of this tag

vickenliu: based on what you said, i think you don't need to over-write this tag

JGiter (approved):

devops/staging/values.yaml
jrhender: What do you mean @vickenliu ? I think I do need to specify which image version we want to run in the staging env. But maybe I'm missing my own point 😆

JGiter (approved):

PR #671, opened 2024-04-02
chore: configure pdb

michalziobro:


michalziobro:
@jrhender please help with decision how to proceed. Some checks fail because of:

> Branch "DOI-3043-Add-pod-disruption-budget-to-IAM-cache-server-helm" is not allowed to deploy to prod due to environment protection rules.

but comparing master -> release: https://github.com/energywebfoundation/ssi-hub/compare/release...master?expand=1
it has some other changes involved. I only want to update helm chart

jrhender (approved):

jrhender:
@michalziobro 
> but comparing master -> release: https://github.com/energywebfoundation/ssi-hub/compare/release...master?expand=1
it has some other changes involved. I only want to update helm chart

No problem, we can merge as is and ignore the `Validate Helm` check as we are confident in this PR.

PR #672, opened 2024-07-11
Undefined pin job

JGiter:
Fix for `Waiting undefined claim undefined` error 
The reason of the error is that in did.processor data for pin job [it is an object](https://github.com/energywebfoundation/ssi-hub/blob/80855fdd2f60da65d1d6b87c54603e6a2c48704c/src/modules/did/did.processor.ts#L66), while pin.processor [expect a string](https://github.com/energywebfoundation/ssi-hub/blob/80855fdd2f60da65d1d6b87c54603e6a2c48704c/src/modules/ipfs/pin.processor.ts#L53). The fix is to specify type of data

### DIFF
--- a/src/modules/ipfs/ipfs.service.ts
+++ b/src/modules/ipfs/ipfs.service.ts
@@ -4,7 +4,11 @@ import { HttpException, HttpStatus, Injectable } from '@nestjs/common';
 import { CID } from 'multiformats/cid';
 import { InjectQueue } from '@nestjs/bull';
 import { Queue } from 'bull';
-import { PIN_CLAIM_QUEUE_NAME, PIN_CLAIM_JOB_NAME } from './ipfs.types';
+import {
+  PIN_CLAIM_QUEUE_NAME,
+  PIN_CLAIM_JOB_NAME,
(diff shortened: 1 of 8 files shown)


src/modules/ipfs/ipfs.service.ts
```diff
@@ -64,10 +68,7 @@ export class IPFSService {
     }
 
     try {
-      await this.pinsQueue.add(
-        PIN_CLAIM_JOB_NAME,
-        JSON.stringify({ cid, claim })
```
jrhender, line 69: Do you recall why `stringify` was used initially? From what I can from the [NestJs Bull queue docs](https://docs.nestjs.com/techniques/queues) it seems fine to use a non-stringified object, but it's not clear to me why we didn't use the object directly in the first place 🤔

jrhender (approved):
Seems okay to me 👍

src/modules/ipfs/ipfs.service.ts
JGiter: No, I don't really remember reason

PR #674, opened 2024-11-06
feat/incremental document sync

JGiter:
* Documents will be synchronized starting from the block of latest synchronization
* Only documents updated since the latest synchronization will be synchronized
* Fixed using of DID synchronization job

### DIFF
--- a/src/modules/did/did.processor.ts
+++ b/src/modules/did/did.processor.ts
@@ -20,6 +20,7 @@ import { DIDDocumentEntity } from './did.entity';
 import { DIDService } from './did.service';
 import {
   ADD_DID_DOC_JOB_NAME,
+  UpdateDocumentJobData,
   UPDATE_DID_DOC_JOB_NAME,
   UPDATE_DOCUMENT_QUEUE_NAME,
 } from './did.types';
@@ -42,28 +43,28 @@ export class DIDProcessor {
   }
 
   @OnQueueActive()
-  onActive(job: Job) {
-    this.logger.debug(`Starting ${job.name} document ${job.data}`);
+  onActive(job: Job<UpdateDocumentJobData>) {
+    this.logger.debug(`Starting ${job.name} document ${job.data.did}`);
   }
 
   @OnQueueStalled()
-  onStalled(job: Job) {
-    this.logger.debug(`Stalled ${job.name} document ${job.data}`);
+  onStalled(job: Job<UpdateDocumentJobData>) {
+    this.logger.debug(`Stalled ${job.name} document ${job.data.did}`);
   }
 
   @OnQueueFailed()
-  onFailed(job: Job) {
-    this.logger.debug(`Failed ${job.name} document ${job.data}`);
+  onFailed(job: Job<UpdateDocumentJobData>) {
+    this.logger.debug(`Failed ${job.name} document ${job.data.did}`);
   }
 
   @OnQueueWaiting()
-  async OnQueueWaiting(job: Job) {
-    this.logger.debug(`Waiting ${job.name} document ${job.data}`);
+  async OnQueueWaiting(jobId: number) {
+    this.logger.debug(`Waiting job ${jobId}`);
   }
 
   @Process(ADD_DID_DOC_JOB_NAME)
-  public async processDIDDocumentAddition(job: Job<string>) {
-    const doc = await this.didService.addCachedDocument(job.data);
+  public async processDIDDocumentAddition(job: Job<UpdateDocumentJobData>) {
+    const doc = await this.didService.addCachedDocument(job.data.did);
 
     await Promise.all(
       doc.service.map(({ serviceEndpoint }) => {
@@ -75,12 +76,14 @@ export class DIDProcessor {
   }
 
   @Process(UPDATE_DID_DOC_JOB_NAME)
-  public async processDIDDocumentRefresh(job: Job<string>) {
+  public async processDIDDocumentRefresh(job: Job<UpdateDocumentJobData>) {
     let doc: DIDDocumentEntity;
     if (this.configService.get<boolean>('DID_SYNC_MODE_FULL')) {
-      doc = await this.didService.addCachedDocument(job.data, true);
+      doc = await this.didService.addCachedDocument(job.data.did, true);
     } else {
-      doc = await this.didService.incrementalRefreshCachedDocument(job.data);
+      doc = await this.didService.incrementalRefreshCachedDocument(
+        job.data.did
+      );
     }
 
     await Promise.all(
--- a/src/modules/did/did.service.ts
+++ b/src/modules/did/did.service.ts
@@ -35,6 +35,7 @@ import { EthereumDIDRegistry__factory } from '../../ethers/factories/EthereumDID
 import { EthereumDIDRegistry } from '../../ethers/EthereumDIDRegistry';
 import {
   DID,
+  UpdateDocumentJobData,
   UPDATE_DID_DOC_JOB_NAME,
   UPDATE_DOCUMENT_QUEUE_NAME,
 } from './did.types';
@@ -58,7 +59,7 @@ export class DIDService implements OnModuleInit, OnModuleDestroy {
     private readonly schedulerRegistry: SchedulerRegistry,
     private readonly httpService: HttpService,
     @InjectQueue(UPDATE_DOCUMENT_QUEUE_NAME)
-    private readonly didQueue: Queue<string>,
+    private readonly didQueue: Queue<UpdateDocumentJobData>,
     private readonly logger: Logger,
     @InjectRepository(DIDDocumentEntity)
     private readonly didRepository: Repository<DIDDocumentEntity>,
@@ -523,7 +524,7 @@ export class DIDService implements OnModuleInit, OnModuleDestroy {
 
   private async pinDocument(did: string): Promise<void> {
     try {
-      await this.didQueue.add(UPDATE_DID_DOC_JOB_NAME, did);
+      await this.didQueue.add(UPDATE_DID_DOC_JOB_NAME, { did });
     } catch (e) {
       this.logger.warn(
         `Error to add DID synchronization job for document ${did}: ${e}`
--- a/src/modules/did/did.service.ts
+++ b/src/modules/did/did.service.ts
@@ -45,6 +45,7 @@ import { SentryTracingService } from '../sentry/sentry-tracing.service';
 import { isVerifiableCredential } from '@ew-did-registry/credentials-interface';
 import { IPFSService } from '../ipfs/ipfs.service';
 import { inspect } from 'util';
+import { LastDidSync } from './lastDidSync.entity';
 
 @Injectable()
 export class DIDService implements OnModuleInit, OnModuleDestroy {
@@ -64,7 +65,9 @@ export class DIDService implements OnModuleInit, OnModuleDestroy {
     private readonly provider: Provider,
     private readonly sentryTracingService: SentryTracingService,
     @Inject('RegistrySettings') registrySettings: RegistrySettings,
-    private readonly ipfsService: IPFSService
+    private readonly ipfsService: IPFSService,
+    @InjectRepository(LastDidSync)
+    private readonly lastDidSyncRepository: Repository<LastDidSync>
   ) {
     this.logger.setContext(DIDService.name);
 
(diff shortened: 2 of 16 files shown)


src/modules/did/did.processor.ts
```diff
@@ -42,28 +43,28 @@ export class DIDProcessor {
...
-  onFailed(job: Job) {
-    this.logger.debug(`Failed ${job.name} document ${job.data}`);
+  onFailed(job: Job<UpdateDocumentJobData>) {
+    this.logger.debug(`Failed ${job.name} document ${job.data.did}`);
   }
 
   @OnQueueWaiting()
-  async OnQueueWaiting(job: Job) {
-    this.logger.debug(`Waiting ${job.name} document ${job.data}`);
+  async OnQueueWaiting(jobId: number) {
```
JGiter, line 61: Argument to this hook differs from others https://docs.nestjs.com/techniques/queues#event-listeners-1

src/modules/did/did.service.ts
```diff
@@ -358,10 +363,30 @@ export class DIDService implements OnModuleInit, OnModuleDestroy {
...
+    ).filter((doc) => {
+      const identity = doc.id.split(':')[3];
+      return changedIdentities.includes(identity);
+    });
+    didsToSynchronize.forEach(async (did) => {
+      this.logger.debug(`Synchronizing DID ${did.id}`);
       await this.pinDocument(did.id);
     });
+
+    await this.latestDidSyncRepository.save({ block: topBlock });
```
jrhender, line 389: Hey @JGiter, thanks for the PR. Really exciting that this code is being improved. 

I think this approach is flawed however. The issue that I see is that jobs are persisted in separate storage (Redis) and then processed asynchronously, but the `latestDidSync` `topBlock` is persisted synchronously. So we could have a situation like this:
1. A bunch of DID update jobs are queued and `lastestDidSync.block` is updated
2. The Redis storage fails or is wiped (it's in-memory, so definitely could happen) -> then the system won't update the DIDs in the lost jobs because `lastestDidSync.block` already ahead of these events.

I'm thinking an alternative approach to get around this problem would be to:
1. Mark updated DIDs as "stale":
  1. Query all of the "changedIdentities" since the last check (as you've done in this PR)
  2. Mark all of these DIDs as "invalid" (in the Postgres Entities). I think something [like this](https://stackoverflow.com/a/62125466) could be used to bulk update entities.
  3. Update `latestDidSync.block`
2. Query all of DID entities with invalid status and add a job to the queue (if it doesn't exist already) 
3. When done processing the job, update the DID to be "valid"

In this above, even if the job queue is reset/wiped, the DID entities will still have invalid status, so job for them can be re-added.

Also, we can change the `GET Did Doc` endpoint to synchronously query the RPC if the cached DID is invalid.

src/modules/did/did.service.ts
```diff
@@ -358,10 +363,30 @@ export class DIDService implements OnModuleInit, OnModuleDestroy {
...
+      `Getting DID update events from block ${fromBlock} to block ${topBlock}...`
+    );
+    const changedIdentities = await this.getChangedIdentities(
+      fromBlock,
+      topBlock
+    );
+    const didsToSynchronize = (
+      await this.didRepository.find({ select: ['id'] })
+    ).filter((doc) => {
+      const identity = doc.id.split(':')[3];
```
jrhender, line 381: I think there might be a helpful function for this in one of our libraries. Something like `getAddress` perhaps?

src/modules/did/did.service.ts
```diff
@@ -486,7 +511,7 @@ export class DIDService implements OnModuleInit, OnModuleDestroy {
 
   private async pinDocument(did: string): Promise<void> {
     try {
-      await this.didQueue.add(UPDATE_DID_DOC_JOB_NAME, did);
+      await this.didQueue.add(UPDATE_DID_DOC_JOB_NAME, { did });
```
jrhender, line 514: @JGiter does this change relate to this comment you made in the PR description?
> Fixed using of DID synchronization job

Does the data need to be in an object?

jrhender (changes requested):
Good start to the PR but I think the issue with the jobs disappearing should be addressed

src/modules/did/did.service.ts
JGiter: Thanks. This is really good improvement

src/modules/did/did.service.ts
JGiter: I can not say why, but when passing job as a string, I observed multiple errors `error parsing JSON` though there are no seemingly JSON parsing in DidProcessor. So I decided to make job as an object, same as in PinProcessor.

src/modules/did/did.service.ts
```diff
@@ -486,7 +508,7 @@ export class DIDService implements OnModuleInit, OnModuleDestroy {
 
   private async pinDocument(did: string): Promise<void> {
     try {
-      await this.didQueue.add(UPDATE_DID_DOC_JOB_NAME, did);
+      await this.didQueue.add(UPDATE_DID_DOC_JOB_NAME, { did }, { jobId: did });
```
JGiter, line 511: Specifying job id to avoid adding job for the same document

src/modules/did/didSyncStatus.entity.ts
```diff
@@ -0,0 +1,26 @@
...
+} from 'typeorm';
+import { DIDDocumentEntity } from './did.entity';
+import { DidSyncStatus } from './did.types';
+
+@Entity()
+export class DidSyncStatusEntity {
+  @PrimaryGeneratedColumn()
+  id: number;
+
+  @OneToOne(() => DIDDocumentEntity, (document) => document.id)
```
JGiter, line 17: Addinig relation only on DidSyncStatusEntity side. The DidDocumentEntity will not be changed

src/modules/did/didSyncStatus.entity.ts
```diff
@@ -0,0 +1,26 @@
...
+  @JoinColumn()
+  document: string;
```
jrhender, line 19: It seems wrong to me that this property is `string` type. I think that typically in a one-to-one join column, the join property has the same type as the joined entity.
See the `Profile` example at the start of the TypeORM documentation: https://orkhan.gitbook.io/typeorm/docs/one-to-one-relations
```suggestion
  @OneToOne(() => DIDDocumentEntity, (document) => document.id)
  @JoinColumn()
  document: DIDDocumentEntity;
```

src/modules/did/did.service.ts
```diff
@@ -215,7 +225,12 @@ export class DIDService implements OnModuleInit, OnModuleDestroy {
         logs: JSON.stringify(logs),
       });
 
-      return this.didRepository.save(updatedEntity);
+      const updated = await this.didRepository.save(updatedEntity);
+      await this.didSyncStatusRepository.save({
+        document: did,
+        status: DidSyncStatus.Synced,
+      });
```
jrhender, line 232: I think this should be in a transaction as I think an invariant should be that there is always a `didSyncStatus` entity if there is a `did` entity.

src/modules/did/did.service.ts
```diff
@@ -495,4 +517,71 @@ export class DIDService implements OnModuleInit, OnModuleDestroy {
...
+      const identity = addressOf(doc.id);
+      return changedIdentities.includes(identity);
+    });
+    await this.didSyncStatusRepository
+      .createQueryBuilder()
+      .useTransaction(true)
+      .update(DidSyncStatusEntity)
+      .set({ status: DidSyncStatus.Stale })
+      .where({ document: In(staleDIDs) })
+      .execute();
```
jrhender, line 583: I seems to me like a better approach would be to rely on the database to do the filtering in the `where` clause, rather than reading all dids into application memory and then filtering in the application. I think that, if the number of DIDs became large enough, this `this.didRepository.find({ select: ['id']` query would probably fail.

I suppose the challenge might be how to use the addresses returned from `changedIdentities` to find the `didSyncStatus` entities to update. I think a couple options are:
1. Store the address of the identity on the `DidSyncStatusEntity`
2. Join to `DIDDocumentEntity` in the query and filter on the `id` property. Of course the addresses would need to be transformed to `did:ethr` DID ids.

Note that I think that option 2 might be slightly less complicated if we got rid of the one-to-one join entity and just added a new column to the DidDocumentEntity.
I think we only really need a single `isStale` boolean column (or could also keep the enum), so the impact on the performance of the database query when reading the DID Document would be  negligible, I think.

jrhender, line 583: Would this query work for existing cached documents (already existing in the cache I mean)? Maybe I missed it, but I don't see a migration that adds a `DidSyncStatusEntity` for all existing cached DID Documents.

jrhender (commented):
Some further comments but I think it is continuing to move in a good direction 👍

src/modules/did/didSyncStatus.entity.ts
JGiter: Yes, I overlooked that

src/modules/did/did.service.ts
JGiter: Thanks for suggestions. Option 2 seems more safe to me, because it excludes case when two status entities connected to the same document. I would also like to avoid changing document entity. It is used everywhere in ssi-hub and I am bit afraid to cause some unexpected changes. And besides property `isState` is not inherently part of the document.

JGiter: Yes, I missed that

src/modules/ipfs/pin.processor.ts
```diff
@@ -56,14 +54,29 @@ export class PinProcessor {
...
+        }
+      }
+      try {
+        const clusterCid = await this.didStoreCluster.save(claim);
+        this.logger.debug(`CID ${cid} saved in cluster by CID ${clusterCid}`);
+      } catch (e) {
+        this.logger.error(
+          `Can not save claim in IPFS cluster. CID ${cid}: ${e}`
+        );
+        return;
```
jrhender, line 78: I think this change is okay to keep in this PR, but is it fair to say that it is somewhat separate from the main changes to the "incremental" did document syncing?

src/modules/did/did.service.ts
```diff
@@ -495,4 +559,102 @@ export class DIDService implements OnModuleInit, OnModuleDestroy {
...
+        const intervalEvents = (
+          await Promise.all(
+            didEventFilters.map((filter) =>
+              this.didRegistry.queryFilter(filter, fromBlock, toBlock)
+            )
+          )
+        ).flat();
+        this.logger.debug(
+          `Fetched ${intervalEvents.length} DID events from interval [${fromBlock}, ${toBlock}]`
+        );
```
jrhender, line 594: Relating to this log line, I've opened this PR: https://github.com/energywebfoundation/ssi-hub/pull/675

src/modules/did/did.service.ts
```diff
@@ -495,4 +559,102 @@ export class DIDService implements OnModuleInit, OnModuleDestroy {
       this.logger.debug(inspect(jobsCounts, { depth: 2, colors: true }));
     }
   }
+
+  /**
+   * Finds DID's changed in given block interval. If fetching of DID's fails at some block, then further fetching stops.
```
jrhender, line 564: Very minor nit, but I wouldn't us the plural rather than possessive "s" here.
```suggestion
   * Finds DIDs changed in given block interval. If fetching of DIDs fails at some block, then further fetching stops.
```

jrhender (commented):
The PR is looking good @JGiter . I'll review a bit more and I'm doing a bit of local testing before we merge, which I should be able to have done by early next week.

src/modules/ipfs/pin.processor.ts
JGiter: That's fair. I moved this error handling in another PR https://github.com/energywebfoundation/ssi-hub/pull/676

jrhender:
Hey @kim-energyweb @hejkerooo @vickenliu , I think this would be a good PR for you to digest if we are continuing to use ssi-hub. The current DID sync approach isn't scalable as the server is re-fetching all DIDs every time there is a sync, so as the total number of tracked DIDs increases eventually the server would be able to complete the fetch in a timely fashion. I've described this issue here: https://energyweb.atlassian.net/wiki/spaces/MYEN/pages/4133978135/SSI-Hub+DID+Caching+Improvement

@JGiter went ahead and took the initiative to implement option number 2 from the doc above; that's this PR. I think it's looking pretty good but I think we should improve the e2e tests for various scenarios prior to merging this. For example, the e2e test should cover the case where the server is tracking a DID, it misses a DID doc update event, and then the sync retrieves the update.

jrhender (approved):
Approving as I don't want my "request review" to block merging. That said, from my perspective, this PR is still untested. @JGiter , IMO, if you're able to add make the automated tests more robust or document the manual testing that you've done, then I think this could be merged. But I would suggestion @kim-energyweb or @hejkerooo  to review at least the high level design before moving the DID sync code in this direction.

PR #677, opened 2025-01-15
feat(ipfs.service): add timeout to get

jrhender:
This PR adds a timeout to the ipfs.service `get`

Todo:
- [x] Decide if return value needs to be change or add catch to did.service
- [x] Test locally


package.json
```diff
@@ -100,6 +100,7 @@
     "didStoreCluster": "npm:@ew-did-registry/did-ipfs-store@0.8.1-alpha.1140.0",
     "didStoreInfura": "npm:@ew-did-registry/did-ipfs-store@0.7.1-alpha.795.0",
     "dotenv": "^16.0.3",
+    "eth-ens-namehash": "^2.0.8",
```
JGiter, line 103: This package is used in ethers/utils.ts, but not installed. Compilation was sucessful, because package was provided by passport-did-auth, but when you install passport-did-auth locally, then compilation fails

jrhender: @JGiter ah, nice find. I had to manually fiddle with the files a bit to get to work but your explanation makes sense and this fix is better.

src/modules/did/did.service.ts
```diff
@@ -322,18 +322,20 @@ export class DIDService implements OnModuleInit, OnModuleDestroy {
...
-        .map(({ serviceEndpoint }) => serviceEndpoint)
-        .filter((endpoint) => IPFSService.isCID(endpoint))
-        .map((cid) => this.ipfsService.get(cid))
-    );
+    return (
+      await Promise.all(
+        service
+          .map(({ serviceEndpoint }) => serviceEndpoint)
+          .filter((endpoint) => IPFSService.isCID(endpoint))
+          .map((cid) => this.ipfsService.get(cid).catch(() => null))
```
jrhender, line 336: @JGiter what do you think of adding a log of the error here?
```suggestion
          .map((cid) => this.ipfsService.get(cid).catch((err) => {
            this.logger.error(`Error resolving ${cid} for ${did}: ${err}`)
            return null
          }))
```

JGiter: It is helpful indeed. I suggest to add it to IpfsService.get though so that we don't log in every place, where getting from IPFS. I added this in next commit

JGiter (approved):

package.json
jrhender: I suppose this was an example of a "phantom" dependency. Perhaps this motivates the use of pnpm in the future: https://rushjs.io/pages/maintainer/package_managers/

JGiter: Yes, perhaps it will be more robast. There is also corresponding eslint rule https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-extraneous-dependencies.md

PR #693, opened 2025-06-11
feat: s3 datasource

azam-ismail:


package.json
```diff
@@ -62,6 +62,7 @@
     "@ew-did-registry/did": "0.8.1-alpha.1140.0",
     "@ew-did-registry/did-document": "0.8.1-alpha.1140.0",
     "@ew-did-registry/did-ethr-resolver": "0.8.1-alpha.1140.0",
```
JGiter, line 64: it would be safer to update other packages as well. They are all built, tested and published together

kim-energyweb (approved):

